PR #1, opened 2011-09-09
18175. don't overwrite existing synced files

paulswartz:
We weren't checking whether the destination existed before copying files, so
we'd inadvertently overwrite existing files.  Now, we check before copying, and
use next_free_filename() to get a free, but related, pathname.


paulswartz:
On IRC:

bendk: looks good to me

Merging.

PR #2, opened 2011-09-09
18188. AttributeError: 'NoneType' object has no attribute 'rfind'

paulswartz:
If we're copying the file, don't bother creating a task to figure out the size;
just grab it from the ItemInfo.


paulswartz:
On IRC:

bendk: seems good to me

Merging.

PR #3, opened 2011-09-14
Albumviewtweaks

bendk:
This is a couple of changes to fix album view bugs.


geoffl:
left one or two comments on things I wasn't too sure about but basically looks good to me.

PR #5, opened 2011-09-15
bz17363: Fix UI feedback when conversion's failed.

geoffl:
When the a conversion has failed for a piece of media, and the user
clicks on 'Clear' in the conversions display it tries to kill the
conversion task.  That's all fine and good except when the task
no longer exists, the kill does not work and throws OSError which is
not caught, and that prevented the cleanup code from executing
which makes subsequent conversions go haywire.


bendk:
merged as e4dc3c2 in master.

PR #6, opened 2011-09-15
bz17973: Avoid IndexError on list pop.

geoffl:
We should check to see whether a list is empty before we call pop() if we
don't intend on catching IndexError.


geoffl:
Can't really explain it but maybe something to do with the play external and/or signal callbacks, and I know play external screws around the code in those places.  Don't really feel like looking at it further as I reckon it'd take too much time, so I added a log message.

Does it look better now?

geoffl:
I added a large comment about this, note, the dbus exception was already caught on object creation, this comment should cover both cases (on create and on use).

anyway, merging

PR #7, opened 2011-09-15
bz18135: Change autodownload order.

geoffl:
The autodownload multi state currently transitions from off to all to new.
It should transition from off to new to all, as transitioning to all
state will make items queued for download which have to be manually canceled.


bendk:
Looks good.  Merged into master as 70f3750.

PR #8, opened 2011-09-15
bz18170: Preference General tab UI rejig.

geoffl:
- Divide the General tab into startup/shutdown and sidebar sections.
- Move show video/audio from podcast prefs from podcast tab to general tab.
- Language with a bold header, and a drop-down.
- Change the explanatory left hand side note to 'Display in:' for choose
  language.
- Move the explanatory note for changing language to precede option to
  make things consistent with other tabs.


bendk:
looks good.  Merged as 81baf37 in master.

PR #9, opened 2011-09-15
bz18122: Make sure we round down in case of odd dimension.

geoffl:
Round down in case of odd dimension handling during conversion, so that we
don't end up with stray black lines around the edge.


bendk:
merged into master as 923ebbc

PR #11, opened 2011-09-15
bz18206: Don't crash when subtitle file fails to copy.

geoffl:
We copy the subtitle file to the target directory of the playing file, but
this can fail for various reasons.  For example, the target could be
read-only, we could have run out of disk space, etc.

Windows port actually caught this error because it would throw an error
when you tried to overwrite an existing file, but we now catch it within the
copy_subtitles_file() function to handle this case for other operating
systems too.


bendk:
merged as 718435e.

PR #12, opened 2011-09-15
Fix for drawing background area on OSX (#18187)

bendk:
The previous code assumed that rectOfColumn_() only included the area that
actually was filled with data.  However, that assumption was plain wrong.


paulswartz:
I'm going to close this pull request since it's not going to work; please re-open after more checkins!

paulswartz:
Closing, since force-pushes don't get picked up in pull requests.

PR #13, opened 2011-09-15
Fix for #18176

bendk:
We weren't including widgetsapi in setup.py because it was only used by
extensions.  Since extensions are dynamically loaded, this messed up the
module finder.  Manually import the module to fix the issue.

### DIFF
--- a/tv/windows/setup.py
+++ b/tv/windows/setup.py
@@ -617,6 +617,21 @@ class bdist_nsis(Command):
             for name in files:
                 self.add_file(os.path.join(root, name))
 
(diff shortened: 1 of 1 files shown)


paulswartz:
I think that the fix for this should be on the py2exe end, rather than the importing end.  That file now throws up a pyflakes error, since it's not used in the file.

bendk:
Definitely a good call paul.  When I was looking it over, I realized that it could be fixed quite cleanly by improving setup.py.

paulswartz:
Much better!

PR #15, opened 2011-09-16
bz18236: Don't crash if we can't lookup a feed id in MultiRowAlbumRendere

bendk:
bz18236: Don't crash if we can't lookup a feed id in MultiRowAlbumRenderer

Added some code to detect that case and return None.  Also added some code to
MultiRowAlbumRenderer to handle no album art.


paulswartz:
Looks good.

PR #16, opened 2011-09-16
Fix for drawing background area on OSX (#18187)

paulswartz:
The previous code assumed that rectOfColumn_() only included the area that
actually was filled with data. However, that assumption was plain wrong.


paulswartz:
Looks good.

PR #18, opened 2011-09-19
Fixes to setup.py for building in Win2008

paulswartz:
- keep out unneeded DLLs
- make sure the path includes DLLs that we do need


bendk:
Code-wise this looks fine, but I get a git conflict when merging to master.  Can you rebase this branch to the latest master?

PR #19, opened 2011-09-20
Bz18239

geoffl:
bz18239: Don't trust the HTTP HEAD response from the server too much.

We use the HTTP HEAD request to get a bit of info about the item before
proceeding to download but it doesn't always work.  Either the server
may not support it, or maybe it it returns some other status code
when the item's really there.

So, if it returned a response code that we did not expect and it was
a HEAD request, try to download anyway and let the download code deal
with failure.

BONUS!  Change a reserved word usage.  type -> typ.

### DIFF
--- a/tv/lib/httpclient.py
+++ b/tv/lib/httpclient.py
@@ -677,15 +677,22 @@ class CurlTransfer(object):
                 self.saw_head_success = True
         elif info['status'] == 401:
             self.handle_http_auth()
-        elif info['status'] in (405, 501):
-            if self.trying_head_request:
-                # server didn't like the HEAD request, so just try the GET
-                self._send_new_request()
-                self.saw_head_success = True
-            else:
-                self.call_errback(UnexpectedStatusCode(info['status']))
         elif info['status'] == 407:
             self.handle_proxy_auth()
+        elif self.trying_head_request:
+            # The response code wasn't what we expected, but we are doing
+            # a HEAD request.
+            #
+            # Servers have "inventive" strategies for dealing with HEAD
+            # requests.  I don't really feel like handling them one by
+            # one since you never know what comes through, so just assume
+            # things are okay and let the actual download handler deal with
+            # failure.
+            #
+            # We handle this before all of the error handling to make sure
+            # we have a chance to catch this.
+            self._send_new_request()
+            self.saw_head_success = True
         elif info['status'] >= 500 and info['status'] < 600:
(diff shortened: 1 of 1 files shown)


kazcw:
This breaks a httpclienttest - the HTTPAuthTest. It looks like behavior has changed for incorrect passwords - this writes an error message to the file where we're expecting it to be left blank.
I don't actually see why this change breaks that particular test, which is a little worrisome. If we're using a HEAD request to test authorization, that makes sense. Otherwise I'd look into why this test and these changes are related.

So this should probably have a couple of changes:
- maintain previous behavior in the auth failure case
- new HEAD test cases that fail with previous behavior and work after this change

geoffl:
Kaz, good catch.  I have updated the changeset so that for proxy auth and for http auth they are caught and handled first.  This makes sense because you want this hack only as a last resort.  With it, it restores the old behavior and the tests pass again.

kazcw:
Would it be practical to write a test case for handling of the funky-status-codes-for-HEAD issue, or is that behavior that's difficult to simulate within our httpclient testing framework?

geoffl:
There's already a test for testing a head which was disallowed--test_head_get_fallback.  The problem in this case the test server throws a 405 (same as the Amazon server), but you can never be sure what you get back with bad servers.  Some return 403 (as was this case for this bug), some return some other codes, default Python http server returns 501.  We can definitely hardcode these codes in the testing server, but I don't think it is practical to hardcode all the list of allowable codes (and possibly disallowable codes too) for testing.  The point, I guess, to make sure here is that there is some test case that we can adequately fall back when the server returns a code when you made HEAD request that would be different from what you would get if you had issued a GET request.  Previously Paul hardcoded handling error 405 in the client for Amazon and the testing server was made to match Amazon, now we handle it for a variety of codes in the client-side as a last resort fallback while keeping a return code of 405 for a disallowed HEAD request.

tv/lib/httpclient.py
```diff
@@ -677,15 +677,22 @@ class CurlTransfer(object):
                 self.saw_head_success = True
         elif info['status'] == 401:
             self.handle_http_auth()
-        elif info['status'] in (405, 501):
```
kazcw, line 680: Removing this block is going to change behavior for non-HEAD 501 errors from UnexpectedStatusCode to PossiblyTemporaryError. This new behavior is probably more correct.

PR #22, opened 2011-09-23
18252. ignore sync of items without video_path

paulswartz:
If for some reason one does make it to conversion_for_info(), throw up a soft
failure.


bendk:
Talked with paul about this one.  It seems good, but it would also be good to try to fix why we are calling conversion_for_info() in the first place.  He's going to work on this one a bit more before we merge it.

paulswartz:
This is ready for review again.

bendk:
Looks great to me.  I think you're the reviewer today, but I guess I'll do the merge.

PR #23, opened 2011-09-27
bz18104.  Catch ValueError in next_free_{filename,directory}().

geoffl:
The next_free_{filename,directory}() functions can throw ValueError if
a set number tries exceed a threshold.  But, code using this does not
catch ValueError, leading to uncaught exception.  Worse, wrap these
around a @returns_filename decorator, which catches ValueError and
reraises ValueError with a different message, so the user/developer
is left confused as to why there is an apparent internal failure with
this suite of functions.

Note, I have been careful to address the following:

(0) Catching ValueError in the case of failure for next_free_filename() or next_free_directory(), though not always returning UI feedback which is obviously not ideal at the moment.  We do log such failures though, hopefully they are not that frequent.

(1) Only proceeding with the requested operation if such temporary filename/next_free_filename()/next_free_directory() succeeded.

(2) next_free_filename() returns a pathname and an open file handle, since this is the only way to ensure safe operation of creating a new file then writing to it (O_CREAT|O_EXCL).  We abuse this API sometimes when we use the pathname component in a file operation (such as a move).  This won't work on Windows as the currently open file will be deleted before the move happens and the deletion won't succeed due to the open file handle.  So as a workaround for the time being, we close it (fp.close()) before any move operation, but obviously this is not ideal.

### DIFF
--- a/tv/lib/conversions.py
+++ b/tv/lib/conversions.py
@@ -533,7 +533,14 @@ class ConversionManager(signals.SignalEmitter):
                 logging.warn("Couldn't find task for key %s", msg['key'])
                 return
             source = task.temp_output_path
-            destination, fp = next_free_filename(task.final_output_path)
+            try:
+                destination, fp = next_free_filename(task.final_output_path)
+                fp.close()
+            except ValueError:
+                logging.warn('_process_message_queue: ' 
+                             'next_free_filename failed.  Candidate = %r',
+                             task.final_output_path)
+                return
             source_info = task.item_info
             conversion_name = task.get_display_name()
             if os.path.exists(source):
@@ -548,7 +555,6 @@ class ConversionManager(signals.SignalEmitter):
             else:
                 task.error = _("Reason unknown--check log")
                 self._notify_tasks_count()
-            fp.close()
             self.emit('task-staged', task)
 
     def _move_finished_file(self, source, destination):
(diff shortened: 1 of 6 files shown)


kazcw:
It seems like when next_free_* runs out of options, the real problem is going to tend to be that we don't actually have permissions to create a file in the directory we're trying to write to (rather than 1000 filenames already being taken). What if we were to put a fallback in next_free_*_candidates that just drops the file into MOVIES_DIRECTORY, rather than have edge-case error handling everywhere we use next_free_*?

geoffl:
Yes, you are definitely right, in fact I think it might be a bit better if we check the errno when we catch the OSError so we can bail out if the error is likely to be permanent.  (To check O_CREAT|O_EXCL we should check for EEXIST and treat the rest as hard errors).

I'm not sure about whether dropping stuff into MOVIES_DIRECTORY is such a good idea though.  The problem is there are places where we abuse these API functions.  I was going to write up something so we can audit the usage of this stuff in the near future but haven't gotten around to it yet.

Any time we need to deal with the external environment it means that the operation may potentially fail.  So calling such functions will mean either having to catch errors or handle a failed return code, in that sense I don't think it's too big a deal.

kazcw:
Ok, well this certainly seems like an improvement.

There's one related issue with the next_free_*_candidates - they can raise errors in two different kinds of situation: got a directory, expected a file / couldn't find a writeable filename. It seems to me that we might generally want to handle these differently: if next_free_filename_candidates receives a directory path, that's a bad input that probably should not have been sent to it in the first place; if it can't find a filename, there's some problem with the directory. Maybe these should be different exceptions (probably subclasses of IOError rather than ValueError), so that we can handle them appropriately.

That could go along with the code cleanup bug for auditing / generally cleaning up usage of these filename utility functions.

PR #24, opened 2011-09-27
bz18288: Correctly pass function arguments.

geoffl:
We save the download status on shutdown.  But due to a TypeError due to
not passing in a required parameter in the function, it gets wiped,
because the restore mechanism falls back to a default due to an
exception being thrown.

This happens with all downloads.

### DIFF
--- a/tv/lib/storedatabase.py
+++ b/tv/lib/storedatabase.py
@@ -639,6 +639,7 @@ class LiveStorage:
                 value = self._converter.from_sql(schema, name, schema_item,
                         value)
             except StandardError:
+                logging.exception('self._converter.from_sql failed.')
                 handler = self._converter.get_malformed_data_handler(schema,
                         name, schema_item, value)
                 if handler is None:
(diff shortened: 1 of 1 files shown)


kazcw:
This has been broken since d5831a4, but it seems we silently catch the TypeError and continue. It looks like we're catching it at _restore_object_from_row, but that would print a warning message that doesn't seem to be showing up and re-raise the exception, so are we catching it and ignoring it farther down the call stack somewhere? It seems like we should at least be logging a warning when something like this happens.

geoffl:
Kaz, I think you are right, we should print a warning if it didn't work. The util.chatter thing is for testing I believe which is why it did not print.

kazcw:
Well miro.util has chatter = True; it looks like what's happening is in _restore_object_from_row, if there's a malformed data handler defined for the field in the object's schema, as long as the malformed data handler succeeds we use its output and move on. I wonder if there are other problems the malformed data handlers are also silently "fixing"?

geoffl:
not sure for the others but for this particular problem I don't think there are other places that it is silently fixing because that's the only place where the dictionary converter is used.

but obviously, a variety of other errors can also be thrown, but with logging.exception we'll be able to catch it as it happens.

kazcw:
Yeah, I just mean in general it seems kind of dangerous that we run the malformed data handlers (which in theory should only run very rarely, right?) without any kind of warning; that could be masking all sorts of problems like this. I've made a branch that prints a detailed warning whenever we handle_malformed_data, so I'll pull-request that tomorrow when it's not my own peer-review day.

PR #25, opened 2011-09-28
include echonest library

kazcw:
- adds pyechonest-4.2.16 in tv/contrib/pyechonest
- adds miro.contrib package to setup.py for each plat

This adds the pyechonest library, a dependency for echonest/echoprint support, by including the pyechonest package in tv/contrib/.

### DIFF
--- a/tv/contrib/pyechonest/__init__.py
+++ b/tv/contrib/pyechonest/__init__.py
@@ -1,0 +1,9 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+"""
+Copyright (c) 2010 The Echo Nest. All rights reserved.
+Created by Tyler Williams on 2009-06-25.
+"""
+
+__all__ = ['config', 'util', 'proxies', 'artist', 'catalog', 'song', 'track', 'playlist']
--- a/tv/contrib/pyechonest/artist.py
+++ b/tv/contrib/pyechonest/artist.py
@@ -1,0 +1,1015 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+"""
+Copyright (c) 2010 The Echo Nest. All rights reserved.
+Created by Tyler Williams on 2010-04-25.
+
+The Artist module loosely covers http://developer.echonest.com/docs/v4/artist.html
+Refer to the official api documentation if you are unsure about something.
+"""
+import util
+from proxies import ArtistProxy, ResultList
+from song import Song
+
+
+class Artist(ArtistProxy):
+    """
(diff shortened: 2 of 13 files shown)


paulswartz:
Should tv/contrib really be a package, instead of something that we add to sys.path?

geoffl:
This looks okay to me, but Paul's suggestion might be a little bit better when doing maintenance work in contrib.

Even though I'm pretty sure we've talked about this before, one thing that I am slightly unsure about is whether it makes some sense to package pyechonest as a separate package as it seems pretty big, the only platform that would be affected here would be Linux since we'd have to bundle it for Windows/Mac anyway.  At least distro packagers would notice this and would have to create a new package to satisfy a dependency (and we wouldn't have to maintain this ourselves) But I guess contrib is meant to satisfy such chicken-egg issues so I guess this would be okay too.

kazcw:
So, my thinking is that if we put pyechonest somewhere that's in sys.path, we'd use it with `import pyechonest` - exactly as we would if that we were using a system package - and that that similarity could lead to confusion and strangeness in several situations, like if someone had pyechonest installed for the system and the sys.path was wrong so the system one was running rather than the local one. By importing it as a package and explicitly using our local version, it's always very clear what's going on. What kind of difficulty when doing maintenance work would this contrib package approach cause? Upgrading the version of pyechonest is as just a matter of replacing the contents of contrib/pyechonest/ with the new versions of the files...

bendk:
Personally, I'm on the fence about this one.  I think Kaz is right that it's nice to be expilcit that we are using a package we shipped.  The one downside is if distros do end up including pyechonest, then it's harder for them to patch our packages.  If we added the contrib dir to sys.path, then they could just delete the directory and we would pick up the system one.

paulswartz:
Being explicit to who? It's not a user-visible difference, and if it becomes an issue for bug reports, exceptions will have the path to the pyechonest library and we'll be able to see if it's ours.

bendk:
I just meant for developers using the library.  But I think paul is still right.  Let's just put the contrib directory at the top of sys.path and use "import pyechonest".  I think if we do that, then we can be very certain that we will get our version of echonest.

kazcw:
Updated according to Paul's suggestion.

paulswartz:
You've now got some extra CONTRIB_DIR lines in setup.py, and there's still an **init**.py in tv/contrib/.  After those changes it looks good to merge.

kazcw:
Removed the **init**.py, refactored away usage of contrib_dir.

PR #26, opened 2011-09-30
Windows Overhaul: GTK 2.24, GStreaber OSS Build 10.7 beta, XULRunner 7.0

bendk:
I think this is finally ready to be really pulled in.  It seems pretty stable and usable to me on my VM.  Janet is going to test over the weakend, but I'd love some peer review as well.


bendk:
hmm, that last commit was bad, closing this one.

PR #28, opened 2011-10-04
Catch exceptions when calling terminate() (#18172)

bendk:
I think the main way this happens is the process quits between when we check
the returncode attribute and when we call terminate().  In that case, don't
bother to log an error.

### DIFF
--- a/tv/lib/subprocessmanager.py
+++ b/tv/lib/subprocessmanager.py
@@ -381,7 +381,14 @@ class SubprocessManager(object):
         self.thread.join(timeout)
         # If things didn't shutdown, then force them to quit
         if self.process.returncode is None:
-            self.process.terminate()
+            try:
+                self.process.terminate()
+            except OSError, e:
+                # Error on terminate.  Normally we should log an error, except
+                # in the case where the process quit by itself before our
+                # terminate() call (see #18172).
+                if self.process.returncode is None:
+                    logging.exception("error calling terminate()")
         self._cleanup_process()
 
     def _on_thread_quit(self, thread):


tv/lib/subprocessmanager.py
```diff
@@ -381,7 +381,13 @@ class SubprocessManager(object):
         self.thread.join(timeout)
         # If things didn't shutdown, then force them to quit
         if self.process.returncode is None:
```
kazcw, line 383: Wouldn't the pythonic solution be to try calling `terminate()` no matter what, skipping the test that we can't rely on?

tv/lib/subprocessmanager.py
```diff
@@ -381,7 +381,13 @@ class SubprocessManager(object):
...
-            self.process.terminate()
+            try:
+                self.process.terminate()
+            except OSError, e:
+                # Error on terminate.  Check if the process quit before our
+                # terminate() call (see #18172)
+                if self.process.returncode is not None:
+                    logging.exception("error calling terminate()")
```
kazcw: This seems to log the exception iffi the process is no longer running after we `terminate()`. Is the case where the process _does_ continue to run not what we should be logging?

tv/lib/subprocessmanager.py
bendk, line 383: I'm not sure what's more pythonic or not, but I think the check is good.  Most of the time returncode will be None, so we can skip calling terminate altogether.

tv/lib/subprocessmanager.py
bendk: Yup, the if statement is totally backwards.  I'll reverse it.

PR #30, opened 2011-10-07
17913. miro-segmenter.c doesn't compile with libav 0.8.0

paulswartz:
Builds for me on Debian Wheezy, which also has an updated version of libav


paulswartz:
Leave this one open until we get a contributor agreement from Helmut

geoffl:
While the original bug reporter was Helmut, the author of this patch was actually Larry, you might want to ping him instead.

PR #31, opened 2011-10-10
Fix windows build after overhaul merge (#18334)

bendk:
We were installing the gstreamer version of the GTK DLLs instead of the GTK
ones.

I'm not sure why this worked before.  I think we may have been trying to
install both files and small differences changed which one ended up surviving.


paulswartz:
When building the branch on the buildbox:

```
*** finding dlls needed ***
error: libgstreamer-0.10.dll: No such file or directory
```

PR #33, opened 2011-10-12
Hide album view switcher for certain displays (#18230, #18237)

bendk:


geoffl:
looks okay to me.  the sharingcontroller should probably have the album view disabled also as we currently do not fetch cover art for those items.

bendk:
I guess I'll merge this one.  I agree about sharingcontroller, but we can change that later.

PR #34, opened 2011-10-12
Address point (b) of bz17155

geoffl:
```
bz17155: Address part (b) of bz17155.

Items jump around when you click on pause all or resume all or cancel all.
The problem is the commands to the remote downloader for each individual
item is sent individually and the response is also replied to individually
meaning that when the items update they are resorted, so for n
items you have n updates and thus n resorts, making the UI jerky.

It'd be nice to batch all of this stuff up so that it's one single ipc
but it's too hard to fix, so when we use a pause all, resume all or
cancel all, first, disable the resort mechanism on the downloads
controller, then, perform an ipc rendezvous with the downloader when
all the status updates have been sent so we can re-enable the sort.
```

BONUS!

Fix a menu typo and adds a dev method that can help us further simulate backend traffic jams.

### DIFF
--- a/tv/lib/dl_daemon/command.py
+++ b/tv/lib/dl_daemon/command.py
@@ -114,6 +114,23 @@ class UpdateDownloadStatus(Command):
         from miro.downloader import RemoteDownloader
         return RemoteDownloader.update_status(*self.args, **self.kws)
 
+# DownloaderSyncRequest/DownloaderSyncNotify: what are these?
+#
+# They are used to as a notification.  Usage: DownloaderSyncRequest command
+# is sent to the downloader and when it is received the downloader sends
+# DownloaderSyncNotify.  Thus, when a DownloaderSyncNotify is received,
+# you can be sure that all commands for the corresponding DownloaderSyncRequest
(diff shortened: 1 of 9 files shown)


geoffl:
I've updated it so that the functions have been renamed and I've added a comment about how the download sync/notify notification mechanism is supposed to be used, but when I force push the comments that Ben made went lost in space.

geoffl:
I've added some more comments and wrapped resort_on_update around an api.  I am not in favor of duplicating comments across all 3 messages however, so I think this update should be adequate.  I put the comments in the pause all / resume all / cancel all in the frontend about what happens which I think gives a better view of what should happen rather than in the specific message.

Regarding nested sync/notify: these mechanism are really meant to be stateless.  They cannot nest, but you may send 2 sync requests in quick succession, and you will receive 2 notifications from the downloader back-to-back.  (increase reference count on send, then decrease when you receive a reply).  But I don't think it's really necessary to wait until the last instance of the reply before we re-enable the automatic resort mechanism, and in any case this only re-enables the sort mechanism, the sort only happens next time you receive a status update which should give the desired behavior.

PR #35, opened 2011-10-12
Copy all gstreamer DLLs to our dist directory (#18343)

bendk:
My previous DLL change left out this part and py2exe wasn't finding them
automatically because gstreamer uses a lot of dynamic DLL loading.


geoffl:
looks good to me

PR #36, opened 2011-10-12
album view column sort fixes (#18278)

bendk:
Always put watched folders at the bottom of the list.  Compare feed names with
case insensitivity.


geoffl:
looks okay to me

PR #37, opened 2011-10-17
bz11640. download Vimeo HD videos

paulswartz:
This also fixes a bug where no Vimeo videos were downloading, the original
reason I looked into this ticket.


bendk:
Looks good.  Merging.

PR #38, opened 2011-10-17
bz18346: Change disk allocation policy for torrents.

geoffl:
Before we were using allocate policy (which may also be sparse).  But it
may make pause immediately after starting a large torrent unusable, as
the close() of the torrent handle would wait for the file to zero-fill.
It can also make shutdown slow after starting big torrents as when you shut
down you shut down the torrents too and zero-filling large things takes
a lot of time.

We now use the compact mode which results in extra disk i/o and possibly
fragmentation but makes everything else usable again.

### DIFF
--- a/tv/lib/dl_daemon/download.py
+++ b/tv/lib/dl_daemon/download.py
@@ -1084,7 +1084,21 @@ class BTDownloader(BGDownloader):
             params["auto_managed"] = False
             params["paused"] = False
             params["duplicate_is_error"] = True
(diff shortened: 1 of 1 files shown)


bendk:
This one looks good.  The one issue is that it will mess with the code that handles leaving space available on the movies directory drive.

Made http://bugzilla.pculture.org/show_bug.cgi?id=18359 to caputure this issue.  I don't think it makes sense to wait on merging this one, since the performance bug is probably more important.

PR #39, opened 2011-10-17
17544. eMusic download support

paulswartz:
eMusic's .emx files can't be downloaded without various cookies, that we can't necessarily get out of the browser.  Instead, we ask the browser to download the file, and then pass it into Miro.


bendk:
Seems good to me.  Merging

PR #44, opened 2011-10-19
bz18370: Do not import libtorrent at the top of util.py.

geoffl:
Util.py is also used in the metadata extractor.  However, due to the
ways that binaries are compiled on Mac OS X it makes libtorrent.so
unusable for anything else other than the Miro.app bundle itself.

NOTE:

I don't really want to fiddle with the paths and all that now to make everything
work since that's probably going to take too long so a big fat comment it is
for now.

### DIFF
--- a/tv/lib/util.py
+++ b/tv/lib/util.py
@@ -49,10 +49,18 @@ import traceback
 import unicodedata
 import urllib
 
-import libtorrent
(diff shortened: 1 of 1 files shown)


paulswartz:
Looks fine, merging.

PR #45, opened 2011-10-20
bz18373: Error checking on DisplayManager.get_current_display().

geoffl:
DisplayManager.get_current_display() can fail with None so make sure that
we check against that before trying to use it.


paulswartz:
Looks fine, merging.

PR #46, opened 2011-10-24
bz18384: Don't prematurely bail in the transcode sink.

geoffl:
The transcode sink is meant to receive the segmented, transcoded data.
We bailed early when we detected the segmenter's quit, but it could
be the case that there is still some residual data to be received
even after segmenter's determined to have quit.

So, don't bail early.


bendk:
seems good to me.

PR #47, opened 2011-10-24
bz18385: Always invoke trancode cleanup handler.

geoffl:
Always invoke transcode cleanup handler.  It wasn't being properly registered
with libdaap.

Also, make sure that in the transcode cleanup handler, all the cleanup
actions are separable by wrapping them individually in try...except blocks
to avoid bailing early on cleanup should one of the statements fail for some
reason.


bendk:
looks good to me.

PR #48, opened 2011-10-24
Unittest fixes.

bendk:
Fixed test_default_column_widths() after I changed the signature to
get_columns_available().  Changed test_corrupt_display_state() to
test_corrupt_view_state() now that we've moved the columns to the view state
table.

Made test_view_types() test all 3 view type combinations.


bendk:
geoff looked at this and said it looks good so I'm merging.

PR #49, opened 2011-10-25
bz18390: Nuke datastructures.py.

geoffl:
It only contained the Fifo datastructure which can be replaced with
deque().

Also replace the stack object implemented with list() with deque()
as it should be more suitable.


kazcw:
This switch over to the standard collection seems consistent and sound

PR #50, opened 2011-10-26
Bz18381

geoffl:
This mainly fixes this:

```
bz18381: Don't let scanning directory feed block backend.

If a directory is big, scan the directory feed in chunks and don't
let it block the backend.
```

But also fixes this:

```
bz18381: Wrap usage of miro_allfiles() judiciously in idle_iterator().

miro_allfiles() may take a long time so make sure that users are eventually
wrapped around in an idle_iterator().
```

as it is related.

I am pretty sure the first fix is okay but for second fix would appreciate some comments.

### DIFF
--- a/tv/lib/feed.py
+++ b/tv/lib/feed.py
@@ -2180,7 +2180,13 @@ class DirectoryScannerImplBase(FeedImpl):
         # files on the filesystem
         scan_dir = self._scan_dir()
         if fileutil.isdir(scan_dir) and not is_file_bundle(scan_dir):
-            all_files = fileutil.miro_allfiles(scan_dir)
+            THRESHOLD = 128
+            all_files = []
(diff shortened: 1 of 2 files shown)


geoffl:
I've just pushed a new set of changes that takes this approach:
- We refactor so that those functions work as iterators that are used by split_item().  Maybe we change find_new_children() to an iterator that yields paths, and we change _make_new_children() to _make_new_child().  Then split_item() would iterate through the output of find_new_children(), pass it to _make_new_child() and yield.  I think this way only split_item() would need the slightly strange empty yield calls.

which I think is probably the cleanest of the options.  I wanted to group everything so that we can pass it all inside as a list, but this looks cleaner and avoids the yield kung-fu.

bendk:
Looks great, merging

PR #53, opened 2011-10-27
bz18375: Don't check Item.downloader attribute directly.

geoffl:
Item.downloader attribute may not exist.  Use has_downloader() method
to check.

blind fix but I think this should be okay.

BONUS: reverse the order of the logical AND check, I think this makes more sense to me.

### DIFF
--- a/tv/lib/startup.py
+++ b/tv/lib/startup.py
@@ -597,6 +597,6 @@ def reconnect_downloaders():
         downloader_.remove()
     manualItems = item.Item.feed_view(feed.Feed.get_manual_feed().get_id())
     for item_ in manualItems:
-        if item_.downloader is None and item_.__class__ == item.Item:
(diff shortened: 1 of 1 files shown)


bendk:
I think this change is good.  It might even make sense to take it further and add a has_downloader() method to FileItem that always returns False.  Then we wouldn't even need the type check.

However, I'm pretty sure it won't fix #18375.  For that one the object was a RemoteDownolader, which is totally wacky.  I'm pretty sure it's another instance of 2 objects having the same ID.

geoffl:
I think it's pretty wacky, but at least, it should hide the problem?  With the latest change, of the item class is not Item, then the test fails and we don't do the second test (cause of the problem).  If there's a copy of the database it should be pretty easy to work out whether there's a duplicate or not.

PR #54, opened 2011-10-27
bz18413: Check display has type before using.

geoffl:
Use hasattr() to check that display has type before using.  It is not
available for all displays.


bendk:
Seems okay to me.

PR #55, opened 2011-10-27
bz18303: Ditch download of items froms shares on disconnect/quit.

geoffl:
Ditch downloading items from shares when we disconnect from share.
If the operation is user-initiated then ask for confirmation.

If it is out of our control then simply print an informational message
that this may have happened.

If we are quitting, disconnect from any active shares which will
cancel downloads.


tv/lib/frontends/widgets/tablist.py
```diff
@@ -818,8 +820,29 @@ class SharingTabListHandler(object):
 
     def on_hotspot_clicked(self, view, hotspot, iter_):
         if hotspot == 'eject-device':
-            # Don't track this tab anymore for music.
             info = view.model[iter_][0]
+            c = app.sharing_tracker.active_downloads_count(info.id)
+            if c:
+                title = _('Share Eject')
+                description = ngettext(
```
paulswartz: ngettext() takes a singular string and a plural string.  The first one should look something like `'There is still %(count)d active download for this share...'`.

geoffl:
this is not ready yet, I still have some changes to make.

paulswartz:
I'm going to close it, then; we can re-open when it's ready.

PR #57, opened 2011-10-31
Device sync small changes

paulswartz:
Two small changes I made, but that are two small to require new tickets.


bendk:
Looks great.

PR #62, opened 2011-11-01
Albumviewtweaks

bendk:
Mostly this is for #18408, but there's a couple other fixes too.


bendk:
Don't merge this one.  I just realized that I haven't pushed what I thought I was pushing.

PR #67, opened 2011-11-05
bz18447: Ignore AttributeError.

geoffl:
This can happen if the modules are in an inconsistent state during shutdown.
Instead of trying to do an orderly pull-down which isn't necessary,
just ignore it.


bendk:
Seems okay.

PR #70, opened 2011-11-07
add some debugging for EMX/AMZ failure cases

paulswartz:
Doesn't actually fix anything, but should give Janet better logs for me to debug.


bendk:
Just adds logging, I can't see an issue with this.

PR #71, opened 2011-11-08
support video data extracted from audio-ish tags

kazcw:
We successfully extract a lot of info like 'Show' and 'Season' from a lot of
videos, but they come from fields that we expect to be audio - e.g. 'Artist' and
'Album', respectively. This copies these fields from the audio fields to the
video fields for cases in which we don't have data that explicitly matches the
video fields.

Adds:
    - 'show' (from artist)
    - 'episode_id' (from title)
    - 'season_number' (from album)
- Makes 'genre' field available for video displays

Increments DatabaseItemSource.VERSION to make the new data visible.


geoffl:
looks ok to me.

PR #72, opened 2011-11-08
Fix for killing dbus after running the unittests

bendk:
I rewrote the script and split things up rather than having one huge line.
I'm not sure what the problem was before, but it was leaving me with an extra
dbus process each time I ran the unittests.  The new version works for me.


geoffl:
looks okay to me.

PR #73, opened 2011-11-09
Fixes for album view (#18458)

bendk:
Made album art take up exactly 6 rows always.

Bumped min-width up by about 30px.  (I think it's actual 29px because the 1st
part changed the min width code)


bendk:
NPR wants to see this in tonight's nightly and it's not too big.  I'm going to be naughty and merge it myself.

paulswartz:
Post-hoc, looks fine.

PR #74, opened 2011-11-10
bz18422: Index active database objects via (id, table name).

geoffl:
It seems that in very rare circumstances we have seen database object
ids being duplicated across different tables, though we believe that
to be impossible based on code inspection and we cannot come up
with a viable explanation as to why it may happen.

Indexing also by the table name will solve the issue.

Additionally, put in a check in database object **init** to confirm
that this is only run within the backend thread to ensure gated
sequential access to the database id allocator.

NOTE: I have deliberately kept the function names the same as I think
they still make sense.  And I have also consciously kept the id allocator 
on a global basis for now.

### DIFF
--- a/tv/lib/database.py
+++ b/tv/lib/database.py
@@ -144,7 +144,7 @@ class DDBObjectFetcher(ViewObjectFetcher):
         self.klass = klass
 
     def fetch_obj(self, id_):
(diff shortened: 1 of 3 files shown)


bendk:
Looks good to me.  I agree that table name makes more sense than class name.

PR #75, opened 2011-11-11
bz18428. fix M3U downloads

paulswartz:
After the refactoring of how we download stuff from Amazon/eMusic, I forgot to
test the M3U files that are used for previews.


bendk:
Seems good to me.

PR #77, opened 2011-11-16
bz18487: Check validity of watched folder feeds before operating.

geoffl:
There are some delayed calls for watched folders but by the time they
actually run the feeds could be no longer available because maybe
it has been deleted.  In this case make sure we don't aimlessly
contnue, which can lead to ObjectNotFoundError when it fetches
the non-existent watched folder feed.


paulswartz:
Looks good to me.

PR #80, opened 2011-11-17
Don't draw the GTK focus ring in albumview (#18272)

bendk:
The issue is that the focus ring gets drawn over the hybrid column, which
messes things up substantially.  I tried to fix this, but I think the best way
is to just disable it entirely.


paulswartz:
Looks fine to me.

PR #84, opened 2011-11-18
Fixes for album/artist text rendering in album view (#18481)

bendk:
Don't render the text if we have a really small amount of space.  Added 3px
to the album view column min-width.


paulswartz:
Looks fine.

PR #85, opened 2011-11-18
bz18468. don't crash if the device doesn't have settings

paulswartz:
We tried to be more efficient by caching the device's settings dictionary, but
if it didn't exist we'd crash. Now we setup a default dictionary if there
aren't any existing settings.


bendk:
Looks quite safe.

PR #86, opened 2011-11-18
Echoprint

bendk:
This doesn't do much, but it does install echoprint on linux systems for us to use later.

I'm not sure if we're using the best system to distribute the echoprint executable, but it matches our system for miro-segmenter, so I guess that's good.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tv/linux/contrib/echoprint-codegen"]
+	path = tv/linux/contrib/echoprint-codegen
+	url = https://github.com/echonest/echoprint-codegen
--- a/tv/lib/frontends/widgets/menus.py
+++ b/tv/lib/frontends/widgets/menus.py
@@ -32,6 +32,7 @@
 import collections
 import itertools
 import logging
+import subprocess
 
 from miro import app
 from miro import errors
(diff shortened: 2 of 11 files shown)


paulswartz:
1. All the echoprint files say "Copyright (c) All rights reserved", but there's also a LICENSE file which says they're under the Boost license, I think. We should figure out which is correct and make sure it's consistent across the code, and also with the GPL.
2. The new menu code will crash on OS X/Windows since we don't define the echoprint function on either platform.

bendk:
I think we won't be breaking the GPL by using echoprint.  We won't link to echoprint.  We will be distributing the two programs together, and communicating through a pipe, which seems okay to me even if echoprint was totally closed source.  This wikipedia article settled the question for me:
http://en.wikipedia.org/wiki/GNU_General_Public_License#Communicating_and_bundling_with_non-GPL_programs

I'm pretty sure echoprint was released under the boot license based on this blog post: http://blog.echonest.com/post/6824753703/announcing-echoprint.  I agree that the "all right's reserved" language is messy, but I'm not sure what to do about that.

I think it's okay that the menu code crashes on windows and OSX since it's in the dev menu.  Once it stops crashing, then we know that we're bundling echoprint properly.

paulswartz:
It looks like we don't have to include the echoprint source at all.  It's available as a Github repository [https://github.com/echonest/echoprint-codegen], so we can add it as a submodule instead of including the code directly.

paulswartz:
Also, I opened an issue to clarify the license status of the Echoprint files: https://github.com/echonest/echoprint-codegen/issues/30

paulswartz:
Okay, the license is fine.  I think a submodule is probably better than including the code ourselves (keeps it up to date), but I'm fine either way.

bendk:
I love the submodule suggestion and the git code is actually newer than the latest release and simplifies the setup.py script.  It still hasn't been changed in several months, so I think it's stable enough for us.

paulswartz:
Can you update the pull request so that it's not pulling in 8e426086?

PR #88, opened 2011-11-21
bz18401. better check for existing connection to signals

paulswartz:
It appears that `self.waiting` can become empty and have `start_conversion()`
called again.  Instead of checking that, just check if we're currently
connected to the signals.


bendk:
I don't totally understand what's going on, but the code definitely is doing what paul wants it to.

PR #89, opened 2011-11-21
bz18179. more correct checking if syncs are done

paulswartz:
Use `is_finished()` rather than checking `started` directly.


bendk:
Looks good.

PR #93, opened 2011-11-21
bz18453. Vimeo videos are not always downloaded

paulswartz:
With some Vimeo URLs, our regex wasn't getting the Vimeo video ID.  Since the
group was named before, we switch to getting the named group all the time,
instead of the the numbered one.


bendk:
Looks great

PR #94, opened 2011-11-22
bz18510: Make sure that mediatype is unicode.

geoffl:
Make sure mediatype is unicode, or else database validation will fail.


bendk:
Seems good to me.

PR #95, opened 2011-11-22
Don't leave files in the echoprint directory

bendk:
It's annoying because echoprint is a submodule, so it means that that
subrepository is always dirty.  We don't have the ability to add something to
echoprint's .gitignore, so let's move the file to our build directory.

Also cleaned up the setup.py script a touch.  I think this way is much simpler
to use.


bendk:
Don't pull this one yet, I just found an error with it

bendk:
Okay, new commit should be good.

PR #96, opened 2011-11-22
Fix to run gst_extractor correctly on linux.

bendk:
We don't have the app module setup in our subprocess module, so we can't use
the previous approach.  That code is pretty weird anyways, since it's trying
to support multiple renderers which is something we don't need anymore.

Implemented a more direct approach, we just put the logic we want in
plat.utils.run_media_metadata_extractor()


geoffl:
whoups, bad geoff.  This definitely looks good!

PR #97, opened 2011-11-23
bz18510: Change subprocess module innards for proper termination.

geoffl:
Make 2 changes.  Firstly, we run the the message loop dispatch in the
main thread and farm off the handler to a separate thread whereas
previously the job role was of the threads was reversed.

Secondly, don't bother checking the returncode attribute on quit.
If it failed to kill the subprocess because of whatever reason
then just ignore it.

nb: different from the bz18510 that was merged in yesterday, because yesterday I sucked and labeled the wrong bug.


bendk:
Looks good to me.

PR #100, opened 2011-11-24
Don't pack invalid daap properties.

geoffl:
Let's not pack invalid daap properties.  If the struct module cannot
convert it into a binary value then just skip over it.

Also better diagnostics for the do_send_reply() function.  If it fails
let's catch it and print and exception before re-raising it for the
upper layers to handle (which normally means disconnection).


geoffl:
this has been merged but github got confused because apparently it was doing maintenance, so I will close this manually.

PR #103, opened 2011-11-25
bz18513: Check self.movie before using.

geoffl:
Don't use self.movie if it is None - otherwise we crash and burn.

### DIFF
--- a/tv/osx/plat/frontends/widgets/quicktime.py
+++ b/tv/osx/plat/frontends/widgets/quicktime.py
@@ -233,6 +233,8 @@ class Player(player.Player):
         return tracks
 
     def get_enabled_audio_track(self):
+        if not self.movie:
+            return None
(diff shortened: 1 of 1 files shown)


tv/osx/plat/frontends/widgets/quicktime.py
```diff
@@ -233,6 +233,8 @@ class Player(player.Player):
         return tracks
 
     def get_enabled_audio_track(self):
+        if not self.movie:
+            return 0
```
paulswartz: If we don't have a movie, this should probably return `None` instead.

geoffl: I basically agree with Paul's comment that this should be None.  Originally I was worried about returning None because the comments imply None means the item is not playing but it seems here it is okay.

Changed it to make it so.

PR #104, opened 2011-11-28
bz18511. crash when expiring auto-sync items

paulswartz:
We were removing items from the device database twice; clearly that doesn't
work.


bendk:
Looks good.

PR #105, opened 2011-11-28
bz18519. crash when syncing multiple items

paulswartz:
We were iterating over the dictionary keys, then popping the values.  Python
doesn't like it when you change the length of a dictionary during iteration, so
just pull the items out along with the name of the converter.


bendk:
Looks good to me.  My one concern is that we are not deleting the keys anymore, but it looks like we just throw away the whole dict at the end of the method anyways.

PR #106, opened 2011-11-29
bz18517: Don't failed_soft() if we notice the thread was already gone.

geoffl:
We could lose the race by the time we come here and as a result
self.thread is already None.  If it is then it is harmless and so
just return.

I think this basically should do the trick with minimal invasive changes.

### DIFF
--- a/tv/lib/subprocessmanager.py
+++ b/tv/lib/subprocessmanager.py
@@ -393,14 +393,29 @@ class SubprocessManager(object):
     def _on_thread_quit(self, thread):
         """Handle our thread exiting."""
 
-        # Igoner this call if it was queued from while we were in the middle
+        # Ignore this call if it was queued from while we were in the middle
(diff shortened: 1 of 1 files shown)


tv/lib/subprocessmanager.py
```diff
@@ -393,14 +393,29 @@ class SubprocessManager(object):
     def _on_thread_quit(self, thread):
         """Handle our thread exiting."""
 
-        # Igoner this call if it was queued from while we were in the middle
+        # Igonre this call if it was queued from while we were in the middle
```
paulswartz: Closer! Should be "Ignore"

paulswartz:
Other than that, looks good.

tv/lib/subprocessmanager.py
geoffl: hah!  I suck :)

PR #107, opened 2011-11-29
bz18502: Create indices for columns that we use joins on.

geoffl:
Using sqlite3 module from Python means we'd better make sure the
JOIN ON columns have indices or else they can get real slow.


bendk:
Seems okay to me.

PR #108, opened 2011-11-30
bz18524: Fix YouTube flv downloads.

geoffl:
We were being sent a URL with a duplicate itag parameter but if we use
it as is it doesn't work and returns an error.  Manually stripping out
duplicate parameters makes things work again.

### DIFF
--- a/tv/lib/flashscraper.py
+++ b/tv/lib/flashscraper.py
@@ -37,7 +37,7 @@ from miro import httpclient
 import urlparse
 import cgi
 from xml.dom import minidom
-from urllib import unquote_plus
+from urllib import unquote_plus, urlencode
 from miro.util import check_u
 
 def is_maybe_flashscrapable(url):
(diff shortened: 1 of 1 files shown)


tv/lib/flashscraper.py
```diff
@@ -124,11 +124,17 @@ def _youtube_callback_step2(info, video_id, callback):
...
+        fmt_url_map = dict()
+        # strip url= from url=xxxxxx, strip trailer.  Strip duplicate params.
+        for fmt, u in zip(fmt_list, stream_map):
+            parts = unquote_plus(u[4:]).split(';')[0].split('&')
+            candidates = []
+            for p in parts:
+                if p not in candidates:
+                    candidates.append(p)
+            fmt_url_map[fmt] = '&'.join(candidates)
+
```
paulswartz: I thought I commented on this, but I guess I didn't.  I think this is more efficient if you use a set, except that you need the first part to be first, since it contains the URL.  Can you use urlparse/urlunparse to handle that, and just use a set to uniquify the query string?

geoffl: You can't have commented because I only did this a couple of hours ago. :)

I actually thought about using urllib and/or urlparse but the problem is the query string, once you pull it apart I couldn't find anything to put it back together again.  But I looked again and I found something.

paulswartz:
Looks good.

PR #109, opened 2011-11-30
bz18526. don't have progress go above 100%

paulswartz:
Since we're guessing about the sizes of converted items, we can guess a bit low
and overshoot the guess.  At that point, we're pretty close to 100% anyways, so
just crop the value to 100% and move on.


geoffl:
looks good to me.

PR #110, opened 2011-11-30
bz18179. even /more/ correct checking if syncs are done

paulswartz:
Syncs that aren't started also aren't done.  This also includes a bugfix for
new device databases that I didn't feel like making a separate ticket/branch
for.


geoffl:
syncs that aren't started aren't also done - but is_finished() returns True.

Anyways the code looks correct so I think we can merge.

PR #111, opened 2011-12-01
bz18284: Make sure we don't kill pending manual downloads on restart.

geoffl:
Pending manual downloads do not have a downloader associated with it
and when they are restarted they are culled.  We shouldn't do this if
it was marked pending.


bendk:
Wow, that's a bad bug.  Nice catch.

PR #112, opened 2011-12-01
bz18332: Don't pass in unicode to pathname2url().

geoffl:
Don't pass in unicode to pathname2url() as it may produce undefined results!
Always convert to str before using.


bendk:
Seems good.  Maybe we want to raise a TypeError if FilenameType isn't unicode in this case, but I'm not really sure about that.  I'll merge this one and let you think about that.

PR #115, opened 2011-12-02
bz18255. actually delete items we expire

paulswartz:
We were deleting the metadata and subtitles, but not the actual files.


bendk:
What if 2 items share the same downloader?  I don't think this does the right thing in that case.

geoffl:
What is the status of this can this be pulled?

paulswartz:
Going to re-evaluate this after some discussion with BDK.

PR #116, opened 2011-12-05
bz18205. better handling of WNDPROC in DeviceTracker

paulswartz:
There was a race condition when setting up our WNDPROC callback,
where our callback was set up before we had the old WNDPROC stored
to call.  Now, we try to get the WNDPROC callback before setting
up our callback, and logging a warning if that doesn't work.


bendk:
If GetWndProcLong fails, can we use DefWndProc as a fallback?

paulswartz:
Ooh, yeah! Good find.

bendk:
Looks great.

PR #118, opened 2011-12-05
device updates

paulswartz:
This adds support for:
- Kindle and Kindle Fire
- HTC Aria
- Galaxy S2
- LG Optimus S
- Toshiba Thrive


bendk:
Mostly just adding data for devices.  Seems very safe.

PR #120, opened 2011-12-07
bz18536: Rework subprocess scheme to allow setting gstreamer variables e...

geoffl:
...arly.

The gstreamer environment variables needs to be set early to prevent
an import prematurely happening and thus missing out on these
variables, and thus not looking up in these paths.

But setting it in the environ is not that easy, because Windows has
certain limitations about what you can and can't set there, among
other limitations.  So rework the code so that these problems are
not present.


paulswartz:
Comments on this branch attached to an older commit: https://github.com/geoffl/miro/commit/3b42bcef1f6c32aef92e2379c5c5fa48ea9de37c

PR #121, opened 2011-12-09
bz18547: Allow setting item kind from the contextual menu.

geoffl:
Allow for setting the item/media kind from the contextual menu.
Works for both single item and multiple items.

### DIFF
--- a/tv/lib/frontends/widgets/itemcontextmenu.py
+++ b/tv/lib/frontends/widgets/itemcontextmenu.py
@@ -132,6 +132,9 @@ class ItemContextMenuHandler(object):
                 convert_menu = self._make_convert_menu()
                 section.append((_('Convert to...'), convert_menu))
 
+            section.append((_('Set media kind as...'),
+              self._make_edit_metadata_menu()))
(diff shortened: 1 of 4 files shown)


tv/lib/frontends/widgets/itemcontextmenu.py
```diff
@@ -429,6 +434,21 @@ def restart_all():
...
+
+        edit_metadata_menu.append((_('Movie'),
+          lambda: app.widgetapp.set_media_kind(u'movie')))
+        edit_metadata_menu.append((_('Show'),
+          lambda: app.widgetapp.set_media_kind(u'show')))
+        edit_metadata_menu.append((_('Clip'),
+          lambda: app.widgetapp.set_media_kind(u'clip')))
+        edit_metadata_menu.append((_('Podcast'),
+          lambda: app.widgetapp.set_media_kind(u'podcast')))
+
```
paulswartz, line 449: Normally I'd say this is better as a loop, but with only 4 items, I'm okay with it.

PR #123, opened 2011-12-09
bz18542: Change continuous playback to True for video and podcast.

geoffl:
Continuous playback to True by default for video and podcast in addition
to audio which is already set to True.


paulswartz:
Looks fine.

PR #125, opened 2011-12-14
bz18549: Fix crash with tablist iter abuse.

geoffl:
We must be very careful when dealing with iters as using a stale one
can cause the entire program to crash.

Here, we make sure that when we forget the child iters we really forget
all children, and also that we remove them from the view.  As well, we
make sure that we handle no tabs selected last, after all changes have
been committed to the table model, so that the tablist iter
view is consistent.

note: Janet verified this and it was good.

### DIFF
--- a/tv/lib/frontends/widgets/tablist.py
+++ b/tv/lib/frontends/widgets/tablist.py
@@ -305,15 +305,14 @@ class TabList(signals.SignalEmitter):
     def remove(self, name):
         with self.removing():
             iter_ = self.iter_map.pop(name)
-            if name in app.tabs.selected_ids:
-                # hack for 17653: on OS X, deleting the selected tab doesn't
-                # send selection-changed - so if the tab we're about to
-                # delete is selected, explicitly change the selection to
-                # this list's root before removing the tab. This preempts
-                # GTK's handling of this case, which is behaviorally the
-                # same.
-                app.tabs._handle_no_tabs_selected(self)
             self.view.model.remove(iter_)
+        if name in app.tabs.selected_ids:
+            # hack for 17653: on OS X, deleting the selected tab doesn't
+            # send selection-changed - so if the tab(s) we've 
+            # deleted is selected, explicitly change the selection to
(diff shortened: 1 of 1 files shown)


tv/lib/frontends/widgets/tablist.py
```diff
@@ -683,24 +684,32 @@ def remove(self, id_list):
...
+                deleted_ids = self.forget_child_iters(iter_)
                 self.view.model.remove(iter_)
+        if set(deleted_ids + id_list).intersection(set(app.tabs.selected_ids)):
+            # hack for 17653: on OS X, deleting the selected tab doesn't
+            # send selection-changed - so if the tab we're about to
+            # delete is selected, explicitly change the selection to
+            # this list's root before removing the tab. This preempts
+            # GTK's handling of this case, which is behaviorally the
+            # same.
+            app.tabs._handle_no_tabs_selected(self)
```
paulswartz: It seems like this branch would get triggered if any selected tabs get deleted, not just all the selected tabs. Maybe use `issuperset()` instead?

``` python
>>> set([1]).intersection(set([1,2,3]))
set([1])
>>> set([1,2,3]).issuperset(set([1]))
True

```

paulswartz:
Other than the line note, this branch seems fine.

tv/lib/frontends/widgets/tablist.py
geoffl: The code used to be inside the for loop so the code would get triggered each time something's deleted anyway.  Now that it's out of the for loop, I basically agree with the reasoning but probably the selection-changed would need to be emitted (and hence this hack) when we change from multiple selected display to a single display?

Anyway, it's probably better to fix the root cause of the issue and remove this hack, which, actually, was the cause of the crash in the first place, but if this hack works as is, I'd prefer not to think too much about it, this will go when the OS X platform code is fixed.

geoffl: One more thing is I submitted a new changeset anyway to fix some inaccuracies in what it describes but the code remains unchanged.

PR #126, opened 2011-12-14
Fix for shutting down after databate upgrade errors.

bendk:
Noticed this error while coding on the metadatamanager branch.


geoffl:
looks good

PR #127, opened 2011-12-21
bz18285: Don't explicitly call destroy() for GTK widgets.

geoffl:
Just let it garbage collect normally.  Destroy() releases PyGTK references
and may do bad things.

notes:

Ben had a look and he said it was okay.
Janet tested and she said it was okay.

There is some code in there that had a week reference to the persistent windows, but they only get torn down when miro shuts down. Since the reference is weak it probably cannot cause a leak/referencing issue.  If you are shutting down you release all resources anyway.  So I have removed those code since I am not really sure that's required.  If there's a good reason why that code should be there, please speak up.

### DIFF
--- a/tv/lib/frontends/widgets/gtk/video.py
+++ b/tv/lib/frontends/widgets/gtk/video.py
@@ -593,7 +593,6 @@ class VideoPlayer(player.GTKPlayer, VBox):
             self.pack_start(self._video_details)
             main_window.main_vbox.pack_start(main_window.controls_hbox)
 
-            self.overlay.destroy()
             self.overlay = None
 
     def rebuild_video_details(self):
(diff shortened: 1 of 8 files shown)


paulswartz:
I don't know enough about PyGTK to judge whether this will actually solve the issues.  But, Ben and Janet both okayed the change, and the code looks fine to me.  Merging.

tv/lib/frontends/widgets/gtk/video.py
```diff
@@ -593,7 +593,6 @@ def _destroy_overlay(self):
             self.pack_start(self._video_details)
             main_window.main_vbox.pack_start(main_window.controls_hbox)
 
-            self.overlay.destroy()
```
paulswartz, line 596: This is causing a blank window to remain, even if full-screen view is removed.  Perhaps instead of `destroy()`, we should be doing something that hides the widgets?

geoffl, line 596: I think this change is made in error; what we should do is keep the destroy (because this isn't a GTK widget) and set the actual GTK window to None.

PR #128, opened 2011-12-27
bz18285: Re-add the call to self.overlay.destroy().

geoffl:
It is needed because this is a signal emitter not a widget, so calling
destroy() is okay.  In addition the destroy() actually closes the window
which was the source of the regression.

Paul can you give it a spin and let me know how it goes for you?


paulswartz:
Works for me!

PR #131, opened 2011-12-30
Fix wheezy helperscript

willkg:
Wheezy has libboost 1.46.  This fixes the helper script so it works.


paulswartz:
Looks good. Do you think your previous employment covers the contribution, or do we need you to sign a new form?

willkg:
Tried to ping you on irc....

I signed a new form and sent it in yesterday.  I haven't gotten an acknowledgement, yet, but I think I'm all set now.  Ben might be on the legal list, so he should be able to tell you one way or the other.

PR #133, opened 2012-01-03
Fix closing the database upgrade dialog.

bendk:
Calling destroy() wasn't doing anything before.  This code fixes it.  It also
seems to work in the cases where we were calling destroy() after run()
returns, like newfeeddialog.py.

### DIFF
--- a/tv/lib/frontends/widgets/gtk/window.py
+++ b/tv/lib/frontends/widgets/gtk/window.py
@@ -365,7 +365,7 @@ class DialogBase(WindowBase):
         raise NotImplementedError()
 
     def destroy(self):
-        self._window = None
+        self._window.response(gtk.RESPONSE_NONE)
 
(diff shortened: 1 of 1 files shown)


tv/lib/frontends/widgets/gtk/window.py
```diff
@@ -365,7 +365,7 @@ def _run(self):
         raise NotImplementedError()
 
     def destroy(self):
-        self._window = None
+        self._window.response(gtk.RESPONSE_NONE)
```
paulswartz, line 368: Should be also be setting `self.window = None` after the response?

bendk, line 368: no because we will set it to None in the run() method.  But there probably should be a comment.  I also see a couple other changes that are needed, I'm going to close this merge request for now.

PR #135, opened 2012-01-04
Fix for menu updating

bendk:
Update menus once we get the initial list of items and restore the selection
from that.

Removed unused argument to ItemListController.on_selection_changed().


geoffl:
looks okay

PR #136, opened 2012-01-04
Fix for check menus.

bendk:
They weren't getting activated when the user turned them off on GTK.  Moved
the code to only affect radio buttons.


geoffl:
looks okay

PR #138, opened 2012-01-09
Fix for #17539

bendk:
I remember noticing the same thing as kaz.  Just added 1px to line_height, which makes sense to me but I'm not 100% sure if it's correct.  However, nothing seems to break with this change, so I think it's okay to commit to master.


paulswartz:
Looks fine.

PR #140, opened 2012-01-14
allow one click url

axet:
miro missing one click url feature. the one currently present isn't working as expected.

i add miroit url handler for mac osx and to the command line.

having this handler will allow easy to download any content from internet by sharing miroit:http:/youtube/movie urls.


paulswartz:
What does this provide that the subscribe links (http://subscribe.getmiro.com/) don't?

axet:
this one provides x-application-miro kind url.

so, right now I unable to call miro from page.

paulswartz:
That's exactly the problem the subscribe URLs are designed to solve.  For example, this link downloads a YouTube video: http://subscribe.getmiro.com/?type=download&url1=http%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3Dd9NF2edxy-M

If that's not working, it's a bug; instructions for filing a ticket are at: http://develop.participatoryculture.org/index.php/GoodBugReports

axet:
yep it is not working. i ll put comment here:

1) it produce subscribe.miro file (which is no good)
2) miro runs on this file and tyring to subscribe on this single youtube url and fail since it is not a feed.

axet:
os mac

paulswartz:
I'm sorry to hear that it's not working for you.  However, you haven't really given us enough information to fix the issue.  http://develop.participatoryculture.org/index.php/GoodBugReports describes the kind of information that we need in order to help.  If you want to work on it yourself, I'd recommend starting with the code that implements the subscribe URLs: https://github.com/pculture/miro/blob/master/tv/lib/singleclick.py

PR #145, opened 2012-01-17
Improvement to the subprocess test.

bendk:
If a key doesn't exist in the result dict, then log the error better.
Just seeing a KeyError isn't that useful.


geoffl:
looks good

PR #147, opened 2012-01-18
bz18625. magnet downloads already have the right path

paulswartz:


tv/lib/dl_daemon/download.py
```diff
@@ -530,16 +532,16 @@ def pick_initial_filename(self, suffix=".part", torrent=False,
...
+        new_filename = os.path.join(download_dir, filename)
+        if not exists:
+            if is_directory:
+                # if this is a torrent and it's a directory of files, then
+                # we create a temp directory to put the directory of files
+                # in.
+                new_filename = next_free_directory(new_filename)
+            else:
+                new_filename, fp = next_free_filename(new_filename)
+                fp.close()
```
paulswartz, line 549: Maybe this block should verify that the path exists?

bendk, line 549: I think that makes sense

bendk:
Overall I think this fix is good and could be merged.  There's a couple comments open though, so I won't do that quite yet.

tv/lib/dl_daemon/download.py
paulswartz, line 549: Should we throw an error in that case, or just create the path?

UPDATE: decided to log a warning, but create the path

PR #149, opened 2012-01-18
remove default font from gtkrc

paulswartz:


paulswartz:
This is actually bug http://bugzilla.pculture.org/show_bug.cgi?id=18309

PR #150, opened 2012-01-19
Delay starting the sharing subsystem a bit.

geoffl:
Nick asked for this feature to not load the startup too much.


bendk:
looks good.

PR #154, opened 2012-01-23
bz18656. don't rename things with real titles

paulswartz:


geoffl:
looks good, let's merge.

PR #155, opened 2012-01-24
Fixed dumb typo

bendk:


bendk:
I know its bad form to merge your own request, but I think this one is trivial enough to merit it.

PR #156, opened 2012-01-24
bz18664: Make sure that the sharing manager is available before using.

geoffl:
Just a few days ago we put in some code that delayed the start of
the sharing manager and tracker to spread the workout out a little
bit.  Sadly, these two things are used at shutdown to checking if
there are outstanding connections (easy to fix) but also used
to manipulate sharing preferences (harder to address).

Let's put the startup of the sharing subsystem back where it was, since
that should be pretty quick, but track the items a bit later.

### DIFF
--- a/tv/lib/startup.py
+++ b/tv/lib/startup.py
@@ -322,6 +322,11 @@ def finish_startup(obj, thread):
     app.download_state_manager = downloader.DownloadStateManager()
     app.download_state_manager.init_controller()
 
+    # Call this late, after the message handlers have been installed.
+    app.sharing_tracker = sharing.SharingTracker()
+    app.sharing_manager = sharing.SharingManager()
+    app.transcode_manager = transcode.TranscodeManager()
+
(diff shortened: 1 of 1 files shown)


bendk:
I just saw this crash on my linux box.  We should make sure the changeset fixes this one too:

## Exception

Traceback (most recent call last):
  File "/home/ben/miro-git/tv/linux/dist/usr/lib/python2.7/site-packages/miro/frontends/widgets/gtk/threads.py", line 40, in wrap_idle_func
    func(_args, *_kwargs)
  File "/home/ben/miro-git/tv/linux/dist/usr/lib/python2.7/site-packages/miro/plat/frontends/widgets/bonjour.py", line 38, in check_bonjour_install
    if app.sharing_manager.mdns_present or request_count > 0:
AttributeError: 'NoneType' object has no attribute 'mdns_present'

geoffl:
also this code is designed to fix the exception seen above.

bendk:
Okay seems good enough to me.

PR #157, opened 2012-01-24
On new metadata speedup

bendk:
Couple of speedups for item.on_new_metadata().

### DIFF
--- a/tv/lib/item.py
+++ b/tv/lib/item.py
@@ -473,9 +473,9 @@ class Item(DDBObject, iconcache.IconCacheOwnerMixin):
     def after_setup_new(self):
         app.item_info_cache.item_created(self)
 
-    def signal_change(self, needs_save=True):
+    def signal_change(self, needs_save=True, can_change_views=True):
         app.item_info_cache.item_changed(self)
-        DDBObject.signal_change(self, needs_save)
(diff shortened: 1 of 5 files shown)


tv/lib/item.py
```diff
@@ -2616,10 +2616,25 @@ def setup_metadata_manager(cover_art_dir=None, screenshot_dir=None):
     app.local_metadata_manager.connect('new-metadata', on_new_metadata)
 
 def on_new_metadata(metadata_manager, new_metadata):
+    # Get all items that have changed using one query.  This is much faster
+    # than calling items_with_path_view() for each path.
+    path_map = collections.defaultdict(list)
+    all_paths = [filename_to_unicode(p).lower() for p in new_metadata.keys()]
```
paulswartz, line 2622: Do you have to generate a list? Or can you just pass a generator expression to `split_values_for_sqlite`?

bendk, line 2622: split_values_for_sqlite calls len() so yes currently.  We could probably re-write it differently, but the current code seems fast enough right now.  I'm not even sure if it would be faster to avoid the list.

paulswartz, line 2622: Okay.

PR #159, opened 2012-01-25
Fix for 18666

bendk:
The issue was that we weren't checking for _get_tracker_for_path returning
None in file_finished.

Decided that that whole system was not well designed and made
_get_tracker_for_path pass on its KeyErrors instead.

Made trackers be a defaultdict instead of having to manually specify which
file types to create keys for.  This seems safer.


geoffl:
looks good

PR #161, opened 2012-01-27
Fix17804

bendk:
Couple of changes based on my work on #17804.

c6d4e267 is my fix for the bug.

18d30bc9 contains improvements to our restart logic and logging for subprocesses.  I'm on the fence if it's important enough to go into the branch, but I'm leaning towards commiting it.

### DIFF
--- a/tv/lib/subprocessmanager.py
+++ b/tv/lib/subprocessmanager.py
@@ -558,7 +558,7 @@ class SubprocessResponderThread(threading.Thread):
 def subprocess_main():
     """Run loop inside the subprocess."""
     if _on_windows():
-        # One windows, both STDIN and STDOUT get opened as text mode.  This
+        # On windows, both STDIN and STDOUT get opened as text mode.  This
         # can causes all kinds of weirdress when reading from our pipes.
         # (See #17804).  Change the mode to binary for both streams.
         import msvcrt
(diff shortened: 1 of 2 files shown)


tv/lib/subprocessmanager.py
```diff
@@ -531,10 +557,13 @@ def run(self):
 
 def subprocess_main():
     """Run loop inside the subprocess."""
-    # make sure that we are using binary mode for stdout
     if _on_windows():
+        # One windows, both STDIN and STDOUT get opened as text mode.  This
```
paulswartz: "On Windows"

tv/lib/subprocessmanager.py
```diff
@@ -339,6 +351,7 @@ def _start(self):
         self.thread.start()
         # work is all done, do some finishing touches
         self.is_running = True
+        self.sent_quit = False
```
paulswartz, line 354: Shouldn't need to do this again, since it's set in `__init__`.

bendk, line 354: We need to do it here in case we restart the process.  The code in **init** is redundant, but I figured it was nice to have the attribute set at all times.

paulswartz, line 354: If we restart the process, shouldn't `sent_quit` still be False?

bendk, line 354: Yes it is...  I guess I was thinking of if we call shutdown() then start().

paulswartz, line 354: Okay, that makes sense; we probably do it in the tests.

PR #162, opened 2012-01-27
Better handling of Integrity errors while updating metadata

bendk:
This is based off of #18670.  I'm not sure what the root problem is, but this
does 2 things:

1) We now always reset new_metadata.  Before if there was an error,
new_metadata would stay the same and we would try to re-insert the data.

2) Better logging

### DIFF
--- a/tv/lib/metadata.py
+++ b/tv/lib/metadata.py
@@ -1309,14 +1309,21 @@ class MetadataManagerBase(signals.SignalEmitter):
         # when we're running mutagen on a music library, but I think that's to
         # be expected.  It seems fast enough in other cases to me - BDK
         self._update_scheduled = False
+        new_metadata_copy = self.new_metadata
         app.bulk_sql_manager.start()
         try:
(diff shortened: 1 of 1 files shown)


tv/lib/metadata.py
```diff
@@ -1309,14 +1309,21 @@ def _run_updates(self):
         # when we're running mutagen on a music library, but I think that's to
         # be expected.  It seems fast enough in other cases to me - BDK
         self._update_scheduled = False
+        new_metadata_copy = self.new_metadata
```
paulswartz, line 1312: This isn't actually a copy of `self.new_metadata`:

``` python
>>> a = {}
>>> b = a
>>> a[1] = 2
>>> b
{1: 2}
```

paulswartz, line 1312: Oh, I see; it's just a copy so that `self._reset_new_metadata()` doesn't remove the data. Nevermind!

PR #163, opened 2012-01-27
bz18677: Make sure the net_lookup_enabled column is non-null.

bendk:
We were doing it for the item table, but I forget to put the same code for the
metadata_status table.

Adding upgrades is weird now that we have device SQL databases.  I think the
code now should do it.  devicedatabaseupgrade.import_from_json() did the right
thing from the start, so we don't need to run the upgrade for those databases.
Instead we just force the version to be 173 to fix databases for users who
have run the nightlies.


paulswartz:
Were we even setting the schema version in the device databases?

bendk:
We set it when we create a new database to the latest version of the schema.  It doesn't really do anything because we never run upgrade_database, but I figure making the value consistent is good.

PR #165, opened 2012-01-27
bz18672: fix changing titles from the edit item box.

bendk:
We just had to fix the key mismatch between ItemInfo and Item/Metadata.


tv/lib/messagehandler.py
```diff
@@ -1544,6 +1544,9 @@ def handle_edit_items(self, message):
             except database.ObjectNotFoundError:
                 logging.warn("EditItems: Item not found -- %s", id_)
                 continue
+            # ItemInfo uses "name", but the metadat system uses "title" now
```
paulswartz: typo

PR #170, opened 2012-01-31
Subprocess cleanup

bendk:
Bunch of checkins to fix issues with the worker process plus one extra one that removes excess logging by metadata.py.


geoffl:
okay let's push.

PR #175, opened 2012-02-03
Donate Manager implementation.

geoffl:
I've put everything into a donate manager singleton to group things together even though it has stuff both in the backend and frontend, but it's not too bad.  In debug powertoys there's one or two things that can be touched by both the backend and frontend, but I'm just not going to worry about it for now.

### DIFF
--- a/tv/lib/donate.py
+++ b/tv/lib/donate.py
@@ -1,0 +1,227 @@
+# Miro - an RSS based video player application
+# Copyright (C) 2005, 2006, 2007, 2008, 2009, 2010, 2011
+# Participatory Culture Foundation
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+#
+# In addition, as a special exception, the copyright holders give
+# permission to link the code of portions of this program with the OpenSSL
+# library.
+#
+# You must obey the GNU General Public License in all respects for all of
+# the code used other than OpenSSL. If you modify file(s) with this
+# exception, you may extend this exception to your version of the file(s),
+# but you are not obligated to do so. If you do not wish to do so, delete
+# this exception statement from your version. If you delete this exception
+# statement from all source files in the program, then also delete it here.
+
+"""``miro.donatemanager`` -- functions for handling donation
+"""
+
+import logging
+import time
+
+from miro import app
+from miro import eventloop
+from miro import prefs
+from miro import signals
+
+from miro.frontends.widgets import donate
+
+from miro.plat.frontends.widgets.threads import call_on_ui_thread
+
+class DonateManager(object):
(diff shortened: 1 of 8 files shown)


tv/lib/donate.py
```diff
@@ -0,0 +1,192 @@
...
+from miro import signals
+
+from miro.frontends.widgets import donate
+
+from miro.plat.frontends.widgets.threads import call_on_ui_thread
+
+class DonateManager(object):
+    NOTHANKS_MULTIPLIER = [50, 100]
+    URL_PATTERN = 'http://getmiro.com/give/m%s/'
+    PAYMENT_URL = 'https://www.stayclassy.org/checkout/donation?cid=7214'
```
paulswartz: Perhaps these should be in appconfig?

tv/lib/donate.py
```diff
@@ -0,0 +1,192 @@
...
+from miro import prefs
+from miro import signals
+
+from miro.frontends.widgets import donate
+
+from miro.plat.frontends.widgets.threads import call_on_ui_thread
+
+class DonateManager(object):
+    NOTHANKS_MULTIPLIER = [50, 100]
+    URL_PATTERN = 'http://getmiro.com/give/m%s/'
```
paulswartz: Also, this URL pattern doesn't appear to work.

tv/lib/donate.py
```diff
@@ -0,0 +1,192 @@
...
+    def reset(self):
+        for pref in [prefs.DONATE_NOTHANKS, prefs.LAST_DONATE_TIME,
+                     prefs.DONATE_COUNTER]:
+            app.config.set(pref, pref.default)
+
+    def show_donate(self, url=None, payment_url=None):
+        if not url:
+            args = [1, 2, 3]
+            try:
+                url = self.URL_PATTERN % args[self.donate_nothanks]
```
paulswartz: Is this just `URL_PATTERN % min(3, self.donate_nothanks + 1)` or am I missing something?

tv/lib/donate.py
geoffl: perhaps, but these are read-only anyway so I didn't put it in.  We can do it like the app serial though but not sure if it's worth it.  Either way I don't have strong attachment.  Do you prefer these to be in app config?

tv/lib/donate.py
geoffl: no, that's basically it.  There used to be something which handled another case but is no longer there.  Anyway, try: except: may not be the best way to express it but none of this is very performance critical so it doesn't seem too bad.

Also, what did you mean above by the URL not working?  The web pages are not in place so they won't navigate to the right place.  I think you get redirected to the main page.

tv/lib/donate.py
paulswartz: That's the way we do other "read-only" items; they're at least overridable by the app.config file.

tv/lib/donate.py
paulswartz: Okay.  Is there a way to disable this while the donation pages aren't available? It'll just be confusing if users get a popup of the Miro homepage.

tv/lib/prefs.py
```diff
@@ -62,6 +62,14 @@ def __ne__(self, other):
         return self.key != other.key
 
 # These are normal user preferences.
+DONATE_PAYMENT_URL          = Pref(key='donatePaymentURL',      default='https://www.stayclassy.org/checkout/donation?cid=7214', platformSpecific=False)
+DONATE_URL_TEMPLATE         = Pref(key='donateURLTemplate',     default='http://getmiro.com/give/m%s/', platformSpecific=False)
+DONATE_ASK1                 = Pref(key='donateAsk1',            default=10,    platformSpecific=False)
```
paulswartz, line 67: Can you document what these variables are?

geoffl, line 67: added better annotation in the lib/donate.py file about what these do.

Also fixed an apparent bug where seemingly the donate window could call show() in the backend.

PR #176, opened 2012-02-07
bz18686: Fix AttributeError in DownloadRateRenderer.

geoffl:
This is a typo.  get_info() should really be get_value().


bendk:
Wow, that one was bad on my part.  Thanks for the fix.

PR #177, opened 2012-02-07
bz18691: Don't call clamp_text() on options when creating option widget.

geoffl:
clamp_text() cuts text off so you could end up with 2 different entries
that become the same when clamp_text() is used.  Not only will indexing
into the original list be wrong when you retrieve the index from the option
widget but some options will not appear (because they are the same).

### DIFF
--- a/tv/lib/frontends/widgets/addtoplaylistdialog.py
+++ b/tv/lib/frontends/widgets/addtoplaylistdialog.py
@@ -32,7 +32,7 @@ code for the "Add to Playlist" dialog.
 """
 
 from miro.gtcache import gettext as _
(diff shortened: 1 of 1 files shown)


bendk:
Seems logical, but I wonder why we had clamp_text there in the first place.  What happens if you have a really long playlist name?  Can it break the layout for that widget?

geoffl:
good point! I thought about this and I don't know if people are going to have really long playlists so I didn't bother.

I am putting in a set_size_request() which should stop the widget from going overly big.  300px across.

bendk:
Looks good to me.

PR #179, opened 2012-02-08
bz18716. OSError when we can't write to a drive

paulswartz:


geoffl:
looks good.  I notice in many other places in the code we catch OSError, IOError, should we be catching from the base class instead?  Either way, it looks good I think we can merge it and then fix up the other places later.  Paul can you file a bug?

paulswartz:
http://bugzilla.pculture.org/show_bug.cgi?id=18726

PR #180, opened 2012-02-08
bz18273. scrollbar incorrectly displayed in Album/List views

paulswartz:
This also fixes an issue where the rating column wasn't correctly displaying stars on hover.


geoffl:
looks okay

PR #181, opened 2012-02-08
bz18593: Fix styles for album view on GTK

bendk:
Extended the focus ring hack to change a couple other style properties that
can make things look funny.


geoffl:
looks okay

PR #182, opened 2012-02-08
bz18671: don't run echonest for video files

bendk:
The issue was that we only had one status to track if we weren't going to run
echonest, STATUS_SKIP.  But we need to distinguish between skipping echonest
because the user has the preference turned off and skipping echonest because
its not an audio file.

Added new status to track this.  Added database upgrade that sets audio files
to have this status.  Fixed device database upgrade so that it sets the status
correctly in the first place.


paulswartz:
Geoffrey, is this good to go?

geoffl:
yes, let's merge

PR #183, opened 2012-02-08
bz18229: fixed default images for album view

bendk:
In the standard view code, we center the image and add a background.  We don't
do that in the album view code, so I just created the effect in GIMP and made
us use those images.


geoffl:
looks okay

PR #189, opened 2012-02-10
bz18718: fix race condition in subprocess tests

bendk:
The tests were hanging because sometimes we would quit with
SubprocessResponder.safe_to_skip_add_idle set True.  To avoid this we just
have to construct a fresh responder in tearDown.

### DIFF
--- a/tv/lib/test/framework.py
+++ b/tv/lib/test/framework.py
@@ -328,6 +328,8 @@ class MiroTestCase(unittest.TestCase):
             patcher.stop()
         # shutdown workerprocess if we started it for some reason.
         workerprocess.shutdown()
+        workerprocess._subprocess_manager = \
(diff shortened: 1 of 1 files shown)


tv/lib/test/framework.py
```diff
@@ -328,6 +328,8 @@ def tearDown(self):
             patcher.stop()
         # shutdown workerprocess if we started it for some reason.
         workerprocess.shutdown()
+        workerprocess._subprocess_manager = \
+                workerprocess.WorkerSubprocessManager
```
paulswartz: Shouldn't this be an instance of the class, rather than the class itself?

bendk: whoops!  Not sure what happened with this commit.  It's fixed now.

PR #196, opened 2012-02-13
bz18737. magnet torrents get stuck in starting up

paulswartz:
Magnet links don't appear to work with the compact storage mode.  If we're
downloading a magnet file, fall back to the old allocate storage mode.


bendk:
Looks good.

PR #197, opened 2012-02-13
bz17528. preference panel tabs get translated at import

paulswartz:


bendk:
This feels slightly automagical to me.  Would it be better to make gettext_lazy have a value() method or something like that that did the gettext call?  That way it would be more obvious where the translation is actually happening.

Just a thought though, if you like it this way, let's just merge it.

paulswartz:
I like this way, because the other code doesn't need to know about it (for the most part).  It just pretends that what it has is a string.

PR #199, opened 2012-02-14
Allow users to pass --g-fatal-warnings on the command line

bendk:
Small improvement I had while trying to figure out #17243.


geoffl:
looks good

PR #200, opened 2012-02-15
bz18735: Make sure we keep a valid reference to window for hiding.

geoffl:
will-close runs an arbitrary signal handler which could call destroy()
and nuke the window widget.  Make sure we keep a reference to it so we
can still hide it on return.


bendk:
Seems okay to me.

PR #203, opened 2012-02-15
Updated clean.sh script on linux.

bendk:
Added a bunch of pyrex files that should be cleaned out


geoffl:
looks good.  but I think pyrex should not dump its stuff in the source directories.  Automatically built sources should be in build directory(s).

PR #205, opened 2012-02-16
bz18686: Don't unconditionally display currently playing item in unplaye...

geoffl:
...d view.

Unplayed really means unplayed, so don't show the currently playing item.


bendk:
The ticket says "[displaying the item] should only be the case if the item started as unplayed and hadn't been marked as played."  I don't think we display the item in this case after the change.

geoffl:
The item is only marked as watched on stop, so if the item's unplayed, the unwatched state will be present and hence the filter will let the item pass through.  Empirically this is what I see so I think it's okay?

bendk:
Ahh, that's correct.  Merging.

PR #206, opened 2012-02-16
Fix18717

bendk:
The easiest way to do this is just to stop using git submodules and just checking in the version of the code we want.


paulswartz:
Fine with me.

PR #209, opened 2012-02-17
bz18255. restart helper process if it's holding files

paulswartz:
This only affects Windows; the other platforms aren't affected by this bug.


paulswartz:
I'm not convinced that this is the best way to solve this bug, but I wasn't able to get Gstreamer to release its hold on the file without killing the process.

bendk:
If it works, it works :)

PR #210, opened 2012-02-17
device crashes

paulswartz:
Ticket #s are in the individual commits.


bendk:
All seem good.  I have some code for #18756 that should handle the issue for partially upgraded databases in a different way.

PR #213, opened 2012-02-21
bz18775: Nuke PersistentWindow.

geoffl:
Does not seem to be used anymore.

note: egrep -re '(persistentwindow|PersistentWindow)' lib linux windows

turned up with nothing.


bendk:
Seems good.

PR #215, opened 2012-02-23
Fix18756

bendk:
Janet gives this the thumbs up.  This is ready to merge, assuming the peer-review is okay.

The main point of all of this is to try to make device database handling work better as devices pass between miro versions.  This meant making it so that the current code outputs things in a way that previous versions can deal with.

Also it meant trying to change the metadata/metadata_status schemas in a way so that future versions can change the schema and still work with this version.  I'm thinking the process would go something like this:
- Suppose Miro 6.0 adds a new metadata processor called "foo"
- we add a foo_status column and set it to noneOk=True.  I think this means that older versions can still read/write to the database and when it writes to the database the column is set to NULL.
- if NULL isn't okay, we can add code that runs an upgrade by selecting all the rows where foo_status is NULL.
- We change the value that gets set to finished_status to 2.  This way if miro 6.0 sets finished_status, then Miro 5.0 will know that it doesn't need to run any metadata processors either.  But if Miro 5.0 sets it, then Miro 6.0 will still get a chance to run the new processor.

### DIFF
--- a/tv/lib/devicedatabaseupgrade.py
+++ b/tv/lib/devicedatabaseupgrade.py
@@ -37,6 +37,7 @@ import urllib
 
 from miro import app
 from miro import databaseupgrade
+from miro import item
 from miro import metadata
 from miro import prefs
 from miro import storedatabase
@@ -222,7 +223,7 @@ class _do_import_old_items(object):
         # versions from running the movie data program on them.  This seems
         # the safest option and old versions should still pick up new metadata
         # when newer versions run MDP.
-        old_item['mdp_state'] = 1
+        old_item['mdp_state'] = item.MDP_STATE_RAN
 
     def insert_into_metadata_status(self, path, file_type, finished_status,
                                     mutagen_status, moviedata_status,
--- a/tv/lib/devicedatabaseupgrade.py
+++ b/tv/lib/devicedatabaseupgrade.py
@@ -69,71 +69,68 @@ class _do_import_old_items(object):
     # FIXME: this code is tied to the 5.0 release and may not work for future
     # versions
 
+    # map old MDP states to their new values
+    mdp_state_map = {
+        None : 'N',
+        0 : 'S',
+        1 : 'C',
+        2 : 'F',
+    }
+
+    # list that contains tuples in the form of
+    # (metadata_column_name, device_item_key
+    column_map = [
+        ('duration', 'duration'),
+        ('album', 'album'),
+        ('album_artist', 'album_artist'),
+        ('album_tracks', 'album_tracks'),
+        ('artist', 'artist'),
+        ('screenshot', 'screenshot'),
+        ('drm', 'has_drm'),
+        ('genre', 'genre'),
+        ('title ', 'title'),
+        ('track', 'track'),
+        ('year', 'year'),
+        ('description', 'description'),
+        ('rating', 'rating'),
+        ('show', 'show'),
+        ('episode_id', 'episode_id'),
+        ('episode_number', 'episode_number'),
+        ('season_number', 'season_number'),
+        ('kind', 'kind'),
+    ]
+
+    insert_columns = ['id', 'status_id', 'file_type', 'source', 'priority',
+                      'disabled']
+    for new_name, old_name in column_map:
+        insert_columns.append(new_name)
+
+    # SQL to insert a row into the metadata table
+    metadata_insert_sql = (
+        "INSERT INTO metadata (%s) VALUES (%s)" %
+        (', '.join(insert_columns),
+         ', '.join('?' for i in xrange(len(insert_columns)))))
+
     def __init__(self, cursor, json_db, mount):
         self.cover_art_dir = os.path.join(mount, '.miro', 'cover-art')
         self.net_lookup_enabled = app.config.get(prefs.NET_LOOKUP_BY_DEFAULT)
         self.mount = mount
         self.cursor = cursor
-        # map old MDP states to their new values
-        self.mdp_state_map = {
-            None : 'N',
-            0 : 'S',
-            1 : 'C',
-            2 : 'F',
-        }
-        # list that contains tuples in the form of
-        # (metadata_column_name, device_item_key
-        self.column_map = [
-            ('duration', 'duration'),
-            ('album', 'album'),
-            ('album_artist', 'album_artist'),
-            ('album_tracks', 'album_tracks'),
-            ('artist', 'artist'),
-            ('screenshot', 'screenshot'),
-            ('drm', 'has_drm'),
-            ('genre', 'genre'),
-            ('title ', 'title'),
-            ('track', 'track'),
-            ('year', 'year'),
-            ('description', 'description'),
-            ('rating', 'rating'),
-            ('show', 'show'),
-            ('episode_id', 'episode_id'),
-            ('episode_number', 'episode_number'),
-            ('season_number', 'season_number'),
-            ('kind', 'kind'),
-        ]
-
-        insert_columns = ['id', 'status_id', 'file_type', 'source', 'priority',
-                          'disabled']
-        for new_name, old_name in self.column_map:
-            insert_columns.append(new_name)
-
-        # SQL to insert a row into the metadata table
-        self.metadata_insert_sql = (
-            "INSERT INTO metadata (%s) VALUES (%s)" %
-            (', '.join(insert_columns),
-             ', '.join('?' for i in xrange(len(insert_columns)))))
-
-        # track which paths already have a row in the metadata_status table
-        cursor.execute("SELECT path FROM metadata_status")
-        self.paths_in_metadata_table = set(row[0] for row in cursor)
-
-        # get info about each item on the device
-        self.device_items = []
-        for file_type in (u'audio', u'video', u'other'):
-            if file_type not in json_db:
-                continue
-            for path, data in json_db[file_type].iteritems():
-                if path not in self.paths_in_metadata_table:
-                    self.device_items.append((file_type, path, data))
-
         # track the next id that we should create in the database
         self.id_counter = itertools.count(databaseupgrade.get_next_id(cursor))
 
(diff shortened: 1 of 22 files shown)


tv/lib/devicedatabaseupgrade.py
```diff
@@ -57,182 +61,197 @@ def import_from_json(live_storage, json_db, mount):
...
+            if file_type not in json_db:
+                continue
+            for path, data in json_db[file_type].iteritems():
+                if path not in self.paths_in_metadata_table:
+                    self.device_items.append((file_type, path, data))
+
+        # track the next id that we should create in the database
+        self.id_counter = itertools.count(databaseupgrade.get_next_id(cursor))
+
+    def invoke(self):
```
willkg: You could put the invoke() code in **call** and thus make the class instance callable.

tv/lib/devicedatabaseupgrade.py
```diff
@@ -57,182 +61,197 @@ def import_from_json(live_storage, json_db, mount):
...
-            current_processor = u'movie-data'
+            finished_status = 0
         else:
-            current_processor = None
-        if file_type == u'audio':
-            echonest_status = 'P' # STATUS_SKIP_FROM_PREF
+            finished_status = 1
+
+        if self.net_lookup_enabled:
+            echonest_status = 'N' # STATUS_NOT_RUN
```
willkg: You're better off putting these values into "constants" somewhere.

tv/lib/devices.py
```diff
@@ -1338,11 +1338,20 @@ def load_sqlite_database(mount, json_db, device_size, countdown=0):
...
+        # force the version to match the current schema.  This is a hack to
+        # make databases from the nightlies match the ones from users starting
+        # with 5.0
+        live_storage.set_version(DB_VERSION)
+    elif live_storage.get_version() != DB_VERSION:
+        # we don't support upgrading databases because the only ones that need
+        # upgrading are ones that have gone through the nightlys.  Just
+        # re-create the database in this case
+        logging.warn("Reseting device database: %r", mount)
+        live_storage.reset_database(init_db=True)
```
willkg: I'm not sure I understand the implications of this. If the database on the device isn't the right version, then will Miro treat the device as if it were new, go through the device from scratch and rebuild the database on the device? And that could potentially happen every time I upgrade Miro? That's pretty intense, if true.

tv/lib/item.py
```diff
@@ -2427,6 +2427,11 @@ def __init__(self, **kwargs):
         else:
             local_path = None
         self._fix_paths_from_database(kwargs)
+        # set values for attributes used in pre-5.0 databases.
+        self.metadata_version = 5 # version used in 4.0.x
+        self.mdp_state = 1 # RAN stat
```
willkg: This is another magic value that should be put in a "constant" somewhere.

tv/lib/devicedatabaseupgrade.py
bendk: Yeah, I went back and forth about that.  I went with a method name because writing the double parens in _DoImportOldItems()() seemed weird.

tv/lib/devices.py
bendk: Yup that's exactly the implication.  I thought this code would be okay because 4.0.x didn't create databases, so I figured the only database where this would make a differences was ones created with the nightlies.  I just realize that we need to deal with versions after 5.0, and this code will definitely not do the right thing there.

I think I will change != to <, so that we won't erase databases created by newer versions of miro.  At that point it will be pretty ugly, but hopefully good enough to push out the door and we can rethink things for the next version.

tv/lib/devicedatabaseupgrade.py
bendk: Sweet to get some feedback from outside the paid dev team!  Thanks for the eyes.

Ben

On Thu, Feb 23, 2012 at 2:09 PM, Will Kahn-Greene
reply@reply.github.com
wrote:

bendk: After thinking about this some more, I think I'll just put the code in the constructor and rename the class name to _do_import_old_items, so running the code will be just like making a function call.

tv/lib/devicedatabaseupgrade.py
```diff
@@ -57,182 +61,197 @@ def import_from_json(live_storage, json_db, mount):
         if action != storedatabase.LiveStorageErrorHandler.ACTION_START_FRESH:
             logging.warn("Unexpected return value from the error handler for "
                          "a device database: %s" % action)
-        handle_failed_upgrade(live_storage.cursor, json_db)
-    finally:
-        live_storage.cursor.execute("COMMIT TRANSACTION")
 
-def _do_import(cursor, json_db, mount):
+class _DoImportOldItems(object):
```
paulswartz: Why the CamelCase? Alternatively, why the strange class name?

tv/lib/devicedatabaseupgrade.py
```diff
@@ -57,182 +62,196 @@ def import_from_json(live_storage, json_db, mount):
...
-    filenames_seen = set()
-
-    for new_name, old_name in column_map:
-        insert_columns.append(new_name)
-
-    next_id = databaseupgrade.get_next_id(cursor)
-    for file_type, path, old_item in device_items:
+    def __init__(self, cursor, json_db, mount):
+        self.cover_art_dir = os.path.join(mount, '.miro', 'cover-art')
+        self.net_lookup_enabled = app.config.get(prefs.NET_LOOKUP_BY_DEFAULT)
```
paulswartz: This, `column_map`, and `insert_columns` could probably be class attributes.

tv/lib/item.py
```diff
@@ -2458,9 +2467,9 @@ def __init__(self, **kwargs):
         self.__initialized = True
 
     def _fix_paths_from_database(self, data):
-        """Make screenshot_path and cover_art_path the correct type.
+        """Make screenshot and cover_art the correct type.
         """
-        for key in ('screenshot_path', 'cover_art_path'):
```
paulswartz, line 2463: `title_tag` -> `title`?

tv/lib/devicedatabaseupgrade.py
bendk: I think this note got messed up somehow, but it's funny that you and will pick apart different aspects of my function object :)  As of the latest commit the class is called _do_import_old_items()

paulswartz: Yeah, it got moved. I just think that it should either be CamelCase (without the _ prefix) or _under_scored, but not both.

tv/lib/devicedatabaseupgrade.py
```diff
@@ -113,126 +102,177 @@ def _do_import(cursor, json_db, mount):
...
-        next_id += 1
+        self.cursor.execute(self.metadata_insert_sql, values)
 
+        if 'cover_art' in old_item:
+            self.upgrade_cover_art(old_item)
+        # Use the RAN state for all old items.  This will prevent old miro
+        # versions from running the movie data program on them.  This seems
+        # the safest option and old versions should still pick up new metadata
+        # when newer versions run MDP.
+        old_item['mdp_state'] = 1
```
paulswartz: Can this explicitly say MDP_STATUS_RAN since we're defining it?

tv/lib/devicedatabaseupgrade.py
```diff
@@ -113,126 +102,177 @@ def _do_import(cursor, json_db, mount):
...
-               "(id, path, current_processor, mutagen_status, "
-               "moviedata_status, echonest_status, net_lookup_enabled, "
-               "mutagen_thinks_drm, max_entry_priority) "
-               "VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)")
-        cursor.execute(sql, (next_id, path, current_processor, 'S',
-                             moviedata_status, echonest_status, False, has_drm,
-                             OLD_ITEM_PRIORITY))
-        status_id = next_id
-        next_id += 1
+            finished_status = 1
```
paulswartz: What do 0 and 1 mean here?

tv/lib/devicedatabaseupgrade.py
bendk: oops, I'll fix that.

tv/lib/devices.py
```diff
@@ -1338,11 +1338,24 @@ def load_sqlite_database(mount, json_db, device_size, countdown=0):
...
+        # Hack for 5.0.
+        #
+        # We didn't create sqlite databases in 4.0.x, so if there is a
+        # database with an earlier version, we know it was created by a
+        # nightly build.  In that case it's not a huge deal to reset it.
+        #
+        # FIXME: Need to write real upgrade code for post-5.0
+        logging.warn("Reseting device database: %r", mount)
+        live_storage.reset_database(init_schema=True)
+
```
paulswartz: Maybe log a warning for the else case too?

paulswartz:
A couple documentation things, and a rebase (can't be automatically merged) and this should be all set.

tv/lib/devicedatabaseupgrade.py
bendk: true/false.  It's a boolean field

paulswartz: Can it use True/False then? Or something like (moviedata_status != 'N')?

tv/lib/devices.py
bendk: Good point.  The else case is greater than right?  I think if DB_VERSION == get_version() we don't need to log anything.

paulswartz: Ah yes. No need to log for ==.

tv/lib/devicedatabaseupgrade.py
bendk: I was worried if sqlite would convert the values correctly, but I just checked and it does.

bendk:
okay rebase is done and I believe all the changes that paul asked for were made.

PR #218, opened 2012-02-23
bz18670: don't run echonest twice for items.

bendk:
The problem we are trying to solve is this:
- We start echonest lookups for a bunch of items
- We change set_net_lookup to False for all items
- This cancels it for all items that are queued up, but it doesn't cancel it
  for items currently being processed (both ones that we are running
  codegens for and ones that we are doing HTTP requests for).
- Soon after, we change set_net_lookup to True for all items
- For those items that were processing, we now have 2 echonest requests
  going.  Eventually this results in us getting an integrity error when we
  try to add 2 rows to the metadata table with the same path/source

Added a couple pieces of code to deal with this:
- In MetadataManager check if we are expecting metadata for a source and
  don't process it if not.
- In _EchonestProcessor track which paths we are working on and ignore
  results for paths that have been removed.

### DIFF
--- a/tv/lib/metadata.py
+++ b/tv/lib/metadata.py
@@ -185,6 +185,10 @@ class MetadataStatus(database.DDBObject):
         column_name = self._source_name_to_status_column[source_name]
         setattr(self, column_name, value)
 
+    def need_metadata_for_source(self, source_name):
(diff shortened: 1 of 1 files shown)


paulswartz:
Looks good.

PR #221, opened 2012-02-24
Fix18752

bendk:


tv/lib/eventloop.py
```diff
@@ -607,3 +607,41 @@ def queuer(*args, **kwargs):
...
+
+    def call_after_timeout(self, timeout, *args, **kwargs):
+        """Call our function after a timeout."""
+        if self.dc is None:
+            self.dc = add_timeout(timeout, self.call_now, self.name,
+                                  args=args, kwargs=kwargs)
+
+    def call_now(self, *args, **kwargs):
+        """Call our function immediately."""
+        self.dc = None
```
paulswartz: Shouldn't this call `cancel_call`, in case someone calls it directly?

tv/lib/frontends/widgets/prefpanel.py
```diff
@@ -559,26 +563,69 @@ def pack_echonest_section(self, v):
...
+
+        return _("Using lookup data for %(net_lookup_count)s "
+                 "out of %(total_count)s items", {
+                     'net_lookup_count': _net_lookup_counts[0],
+                     'total_count': _net_lookup_counts[1],
+                 })
+
+    def _update_net_lookup_butons(self):
+        """Enable/Disable the buttons for echonest lookup."""
+        global _net_lookup_counts
```
paulswartz, line 612: Technically you don't need this, since you aren't setting `_net_lookup_counts`.

tv/lib/eventloop.py
bendk: Oops, I'll fix that.

tv/lib/frontends/widgets/prefpanel.py
bendk, line 612: Yeah, I like to make it explicit though.

paulswartz:
Looks good.

PR #223, opened 2012-02-27
bz18787. recent FFmpeg doesn't have `-vpre slow`

paulswartz:
Instead, it should be `-preset slow`.


bendk:
Seems like a bit of a hack, but I guess that's what we're have to work with.

PR #225, opened 2012-02-28
bz18795: Update binary kit.

geoffl:
Updates the binary kit.  Corresponding change: update pycurl to a same one but built against a newer version of libcurl which apparently fixes some downloading issues.


bendk:
This one is definitely good.

PR #229, opened 2012-02-28
bz18803: make exceptions derive from StandardError

bendk:
Changed code in a few places to only catch StandardErrors rather than having
separate code for KeyError/SystemError.

I'm not 100% sure that it's a good idea to merge this at this point in the dev cycle, but I'm leaning towards merging it.  Another way to do it would be to only change the base classes, not the except statements.

### DIFF
--- a/tv/lib/coverart.py
+++ b/tv/lib/coverart.py
@@ -42,11 +42,11 @@ from miro import prefs
 from miro import util
 from miro import fileutil
 
-class UnknownImageObjectException(Exception):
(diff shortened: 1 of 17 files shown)


geoffl:
I just went to double check on the Python docs and it said this for Exception:

All built-in, non-system-exiting exceptions are derived from this class. All user-defined exceptions should also be derived from this class.

But subclassing from StandardError means it is easy to have a catch-all to catch our own exceptions in addition to SystemExit and KeyboardInterrupt.

Let's merge.

PR #233, opened 2012-02-29
bz18707. new_search_podcasts don't update immediately and display items

paulswartz:
We weren't sending a change signal, so the notification about the update-in-progress was being delayed until something else happened.


geoffl:
looks good

PR #234, opened 2012-03-01
bz17793: Redirect stderr/stdout on Windows.

geoffl:
Try to write to a place we have permissions to rather than the current
directory of the binary which we might not have access to.

### DIFF
--- a/tv/windows/Miro.py
+++ b/tv/windows/Miro.py
@@ -33,8 +33,29 @@
 import os
 import sys
 import logging
+import tempfile
 
 def startup(argv):
+    # Rewrite the stdout and stderr to catch cases debug being printed via
+    # a print or via a write to stderr.  Normally we would use shortAppName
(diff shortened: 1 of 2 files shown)


bendk:
So duplicating the logging module is definitely bad.  Could we reset sys.stderr back to devnull when we start the logging code?  One more question, what happens in console mode?  We want mirodebug.exe to print stuff to stdout.

geoffl:
okay, I agree that's a better way.  So let's setup a log but set it to devnull when we start the module.

When you run mirodebug.exe with the --unittest flag it runs a different startup() so it should not be an issue.

bendk:
I think mirodebug is also used in the run.sh script.  It's built to run in console mode so that stdout actually does work.  I use it to run miro on windows and see the output of the logs without using tail or something like that.

paulswartz:
Changes look good.

PR #236, opened 2012-03-01
bz18779: Don't translate buttons at import-time.

geoffl:
(This should fix it for the buttons at least, I think).

### DIFF
--- a/tv/lib/dialogs.py
+++ b/tv/lib/dialogs.py
@@ -81,58 +81,63 @@ import threading
 from miro import eventloop
 from miro import signals
 from miro import util
-from miro.gtcache import gettext as _
+from miro.gtcache import gettext_lazy, gettext as _
 
 class DialogButton(object):
     def __init__(self, text):
-        self.text = text
+        self._text = text
     def __eq__(self, other):
         return isinstance(other, DialogButton) and self.text == other.text
     def __str__(self):
         return "DialogButton(%r)" % util.stringify(self.text)
+    @property
+    def text(self):
+        return unicode(self._text)
 
-BUTTON_OK = DialogButton(_("OK"))
-BUTTON_APPLY = DialogButton(_("Apply"))
-BUTTON_CLOSE = DialogButton(_("Close"))
-BUTTON_CANCEL = DialogButton(_("Cancel"))
-BUTTON_DONE = DialogButton(_("Done"))
-BUTTON_YES = DialogButton(_("Yes"))
-BUTTON_NO = DialogButton(_("No"))
-BUTTON_QUIT = DialogButton(_("Quit"))
-BUTTON_CONTINUE = DialogButton(_("Continue"))
-BUTTON_IGNORE = DialogButton(_("Ignore"))
-BUTTON_IMPORT_FILES = DialogButton(_("Import Files"))
-BUTTON_SUBMIT_REPORT = DialogButton(_("Submit Crash Report"))
-BUTTON_MIGRATE = DialogButton(_("Migrate"))
-BUTTON_DONT_MIGRATE = DialogButton(_("Don't Migrate"))
-BUTTON_DOWNLOAD = DialogButton(_("Download"))
-BUTTON_REMOVE_ENTRY = DialogButton(_("Remove Entry"))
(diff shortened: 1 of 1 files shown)


paulswartz:
You can also do this with `gettext_lazy` from utils. Actually, you probably should because otherwise these strings won't get picked up for translation.

geoffl:
changed it to use gettext_lazy() + a run through unicode() to make the strings look right.

geoffl:
I was careful to make the text attribute a function with a property decorator, which should ensure that while the object is created at import time the string is not touched until the .text property is accessed, so I think that should work, but using gettext_lazy() is probably better.

tv/lib/dialogs.py
```diff
@@ -81,58 +81,61 @@ def handlePizzaAnswer(dialog):
 from miro import eventloop
 from miro import signals
 from miro import util
-from miro.gtcache import gettext as _
+from miro.gtcache import gettext_lazy, gettext as _
 
 class DialogButton(object):
     def __init__(self, text):
-        self.text = text
+        self._text = unicode(gettext_lazy(text))
```
bendk: Pretty sure this is equivalent to just gettext(text)

tv/lib/dialogs.py
```diff
@@ -81,58 +81,61 @@ def handlePizzaAnswer(dialog):
...
-BUTTON_STOP_WATCHING = DialogButton(_("Stop Watching"))
-BUTTON_RETRY = DialogButton(_("Retry"))
-BUTTON_START_FRESH = DialogButton(_("Start Fresh"))
-BUTTON_INCLUDE_DATABASE = DialogButton(_("Include Database"))
-BUTTON_DONT_INCLUDE_DATABASE = DialogButton(_("Don't Include Database"))
+    @property
+    def text(self):
+        return _(self._text)
+
+BUTTON_OK = DialogButton("OK")
```
bendk: The point of gettext_lazy is to trick our text extractors into thinking we're running gettext, but not actually run gettext until later.  We need the strings to be inside a _() call so that the translation utils pick them up.

One way to do it is this:
-  import gettext_lazy as _
- Change the constructor code so that it stores its argument in _text instead of text
- Add the text property that returns unicode(self._text).  This is when the translation will actually happen
- Keep the code that creates the buttons the same as before.

tv/lib/dialogs.py
paulswartz: Ben is correct.  `gettext_lazy` is evaluated when `unicode` is called.  You should just be able to replace `_('foo')` with `gettext_lazy('foo')`.

tv/lib/dialogs.py
geoffl: Ok, I am confused.  So the original solution I had does exactly that.  In this case I don't see the need to use gettext_lazy() since we use the text property and we can just use _() as normal, that should be evaluated as required.

paulswartz: 1) You shouldn't need to use `unicode()`; you could just set `self.text` to `gettext_lazy('foo')`
2) Even with this solution, you've removed all the `_()` so the translation-finder doesn't see any of them.

geoffl: latest version should be okay now.

tv/lib/dialogs.py
```diff
@@ -81,58 +81,63 @@ def handlePizzaAnswer(dialog):
...
-BUTTON_DONT_INSTALL = DialogButton(_("Don't Install"))
-BUTTON_SUBSCRIBE = DialogButton(_("Subscribe"))
-BUTTON_STOP_WATCHING = DialogButton(_("Stop Watching"))
-BUTTON_RETRY = DialogButton(_("Retry"))
-BUTTON_START_FRESH = DialogButton(_("Start Fresh"))
-BUTTON_INCLUDE_DATABASE = DialogButton(_("Include Database"))
-BUTTON_DONT_INCLUDE_DATABASE = DialogButton(_("Don't Include Database"))
+    @property
+    def text(self):
+        return unicode(self._text)
```
paulswartz, line 95: I'm still not sure you need this, but it doesn't hurt anything.

bendk:
Looks good now.

PR #237, opened 2012-03-01
bz17749: Re-read Windows proxy configuration on every query.

geoffl:
Re-read the Windows proxy configuration in case it changes.


bendk:
Seems good.

PR #239, opened 2012-03-01
bz4763: Add file filters for the "Open" dialog

bendk:
Big thanks to Ben Sturmfels.

It's nice to fix a 4-digit ticket.

### DIFF
--- a/tv/lib/filetypes.py
+++ b/tv/lib/filetypes.py
@@ -41,6 +41,7 @@ VIDEO_EXTENSIONS = ['.mov', '.wmv', '.mp4', '.m4v', '.ogv', '.anx',
                     '.rmvb', '.mkv', '.m2v', '.ogm', '.webm', '.ogx']
 AUDIO_EXTENSIONS = ['.mp3', '.m4a', '.wma', '.mka', '.flac', '.ogg', '.oga']
 FEED_EXTENSIONS = ['.xml', '.rss', '.atom']
+TORRENT_EXTENSIONS = ['.torrent']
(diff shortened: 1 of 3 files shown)


tv/lib/frontends/widgets/application.py
```diff
@@ -469,7 +469,23 @@ def reveal_conversions_folder(self):
 
     def open_video(self):
         title = _('Open Files...')
-        filenames = dialogs.ask_for_open_pathname(title, select_multiple=True)
+        audio_extensions = [mem.replace(".", "")
```
paulswartz: Probably should be `mem[1:]` for all three.

tv/lib/filetypes.py
```diff
@@ -41,6 +41,7 @@
                     '.rmvb', '.mkv', '.m2v', '.ogm', '.webm', '.ogx']
 AUDIO_EXTENSIONS = ['.mp3', '.m4a', '.wma', '.mka', '.flac', '.ogg', '.oga']
 FEED_EXTENSIONS = ['.xml', '.rss', '.atom']
+TORRENT_EXTENSIONS = ['.torrent']
```
paulswartz, line 44: Don't we have this already?

bendk, line 44: As far as I can tell, no.

tv/lib/frontends/widgets/application.py
bendk: Good point.  I'll fix this.

paulswartz:
Do we have a contributor agreement from Ben? If yes, this is good to merge.

bendk:
contributor agreement is in place.

PR #240, opened 2012-03-01
bz18825: use better image for podcasts by default.

bendk:
I also refactored this code so that it wasn't one huge if statement.


paulswartz:
Looks good.

PR #242, opened 2012-03-02
bz18789: handle unicode data from echonest.

bendk:
The reason this was breaking on osx and not linux/windows has to do with the
fact that we use the json module rather than simplejson.  They have slightly
different ways of handling unicode.

To make things uniform, decode the response ourselves and send the unicode
value to the json module.  This should make the strings returned by the json
module unicode.


paulswartz:
Looks fine.

PR #243, opened 2012-03-05
Fixes for the unittests.

bendk:
Adding the cooldown period to echonest codegen processing is what broke
things.


paulswartz:
Looks fine.

PR #245, opened 2012-03-05
bz18831: fix attribute error during force save db error

bendk:
Implemented the same idea without the mock module because we can't easily
import it on all platforms.


paulswartz:
Oh, I see.  We can't import the 'test' module on Windows/OS X, I guess because it's not packaged with everything else.

PR #248, opened 2012-03-05
bz18705: Change blip.tv to Blip.

geoffl:
I've only changed the display name and kept the internals the same and
things should continue to work as normal.


paulswartz:
Looks fine.

PR #251, opened 2012-03-06
bz18835: better way of converting filenames to ASCII

bendk:
Not sure why this code is necessary, but I believe this new way should accept
any unicode or bytestring and convert it to plain ASCII.


geoffl:
I would just get rid of the ascii conversion.  We should convert it to something sane and then passing it to zipfile, instead of converting to ascii.  Doesn't seem like the right thing to do and will probably fail if you have something in something other than English.  The ascii conversion's been there for years but I don't think that's necessarily the right thing to do.

bendk:
After skimming through the zipfile mailing list, I believe that ASCII is actually the safest thing to use.  There seems to be some support for other things, but not much.

I agree that the conversion is wonky though.  I'm thinking we should switch to tarballs to fix things.

bendk:
I'm going to close this one and open a new pull request.

PR #253, opened 2012-03-06
bz18840: don't run movie data on non-media files.

bendk:
Our gstreamer movie data programs misidentify image files as video because
they can render the image.

This simplest fix is to skip moviedata if mutagen + the extension indicate
that the file is not a media file.  Also if moviedata returns 0 duration, then
assume the file is other even if movie data thinks its a video.


geoffl:
looks good

PR #255, opened 2012-03-07
bz17823. OS X should emit selection-changed on delete

paulswartz:
OS X doesn't emit `selection-changed` signals when calling `reloadData()`.
Instead, we check if the selection is no longer valid and manually send the
signal ourselves.


paulswartz:
Fixed the typo mentioned by Geoff.

bendk:
Seems good enough to me.

PR #256, opened 2012-03-07
bz18751: running echonest on video files.

bendk:
This would happen when mutagen succeeded and moviedata failed.  Added a test
and fix for the issue.


geoffl:
looks good

PR #258, opened 2012-03-07
Fix18844

bendk:
The main part of this is the first commit, but I figured that we might as well combine the gst_extractor code for linux and windows since they were almost identical.


geoffl:
looks good

PR #259, opened 2012-03-08
Supportzipfixes

bendk:


tv/lib/test/utiltest.py
```diff
@@ -908,3 +910,90 @@ def test_subdirs(self):
...
+        open(full_path, "wt").write(contents)
+        if not should_skip:
+            self.correct_files.append(archive_name)
+
+    def check_backup(self):
+        backup = util.SupportDirBackup(self.support_dir, self.skip_dirs,
+                                       max_size=1000000)
+        archive = zipfile.ZipFile(backup.fileobj(), 'r')
+        errors = archive.testzip()
+        if errors is not None:
```
paulswartz: The regular way to write this (in Python < 2.7, at least) is `self.assertTrue(errors is None, msg)`.

tv/lib/test/utiltest.py
```diff
@@ -908,3 +910,90 @@ def test_subdirs(self):
...
+        for i in xrange(10):
+            self.add_file_to_support_dir('big-file-%s' % i,
+                                         contents=large_data)
+        # check that we don't max an archive file too much bigger than our max
+        # size
+        max_size = 1000000 # 1MB
+        backup = util.SupportDirBackup(self.support_dir, self.skip_dirs,
+                                       max_size=max_size)
+        filesize = os.stat(backup.backupfile).st_size
+        if filesize > 1100000:
```
paulswartz: Same with this: `self.assertTrue(filesize < 1100000, msg)`.

tv/lib/util.py
```diff
@@ -1162,3 +1164,77 @@ def split_values_for_sqlite(value_list):
     for start in xrange(0, len(value_list), CHUNK_SIZE):
         yield value_list[start:start+CHUNK_SIZE]
 
+
+class SupportDirBackup(object):
+    """Backup the support directory to send in a crash report."""
+    def __init__(self, support_dir, skip_dirs, max_size):
+        logging.info("Attempting to back up support directory: %r",
+                     support_dir)
+        uniqfn = "%012ddatabasebackup.zip" % random.randrange(0, 999999999999)
```
paulswartz: Maybe do this with `time.time()` or something, so that they're ordered?

tv/lib/util.py
```diff
@@ -1162,3 +1164,77 @@ def split_values_for_sqlite(value_list):
...
+        self.backupfile = os.path.join(tempfile.gettempdir(), uniqfn)
```
paulswartz: Should we check that the file doesn't exist already?

tv/lib/util.py
```diff
@@ -1162,3 +1164,77 @@ def split_values_for_sqlite(value_list):
...
+        archive = zipfile.ZipFile(self.backupfile, "w")
+        self.skip_dirs = [os.path.normpath(d) for d in skip_dirs]
+
+        total_size = 0
+        for root, dummy, files in os.walk(support_dir):
+            if self.should_skip_directory(root):
```
paulswartz: This works, but it might be more efficient if you filtered the unneeded directories from `dirnames` (called `dummy` here).

``` python
remove = [f for f in dummy if self.should_skip_directory(f)]
for f in remove:
    dummy.remove(f)
```

You'd also be able to remove the path normalizing stuff in `should_skip_directory`.

tv/lib/util.py
```diff
@@ -1162,3 +1164,77 @@ def split_values_for_sqlite(value_list):
...
+            if total_size >= max_size:
+                logging.warn("Support directory backup too big.  "
+                             "Quitting after %s bytes", total_size)
+                break
+        archive.close()
+        logging.info("Support directory backed up to %s (%d bytes)",
+                     self.backupfile, os.path.getsize(self.backupfile))
+
+    def should_skip_directory(self, directory):
+        if os.path.islink(directory):
```
paulswartz: `os.walk()` doesn't recurse into links by default, so this isn't needed.

tv/lib/util.py
bendk: I think I'll change it to next_free_filename() so they're ordered and  it matches how we generate other filenames like this.

paulswartz:
Looks good.

PR #260, opened 2012-03-09
bz18853. update device View() call for new API

paulswartz:
f867313892be7c17be858e8421aab2e7929fcb83 changed the View API to take a dbinfo
argument, but this call wasn't updated to grab that field from the view we're
duplicating.


bendk:
Looks great.

PR #261, opened 2012-03-09
bz17549: Disable album view button on sharing controller for now.

geoffl:
Patch from Ben until we properly support album art on sharing.


paulswartz:
Looks fine.

PR #262, opened 2012-03-09
Bz18846: Don't kick up crash dialog in ItemInfoCache.

geoffl:
```
bz18446: Don't kick up crash dialog in ItemInfoCache.

There are some cases where it may be possible for the cache not to have
the item info data because Miro shut down abnormally (e.g. maybe during
a hard crash or loss of power).

Just print a warning in this case to mirror what happened in production.
The crash dialog was only displayed in debug mode anyway.

Secondly, this may be a repeatable error because the recovery code did not
schedule save of the newly created item info to the backing store.  Change it
to mirror what is done for an item info add but without signaling.
```

Bonus!  Includes:

```
Cleanup MiroInterceptorResponder.

Change it to mirror the convention for Objective-C code.
```

### DIFF
--- a/tv/lib/iteminfocache.py
+++ b/tv/lib/iteminfocache.py
@@ -223,11 +223,16 @@ class ItemInfoCache(signals.SignalEmitter):
         try:
             return self.id_to_info[id_]
         except KeyError:
-            app.controller.failed_soft("getting item info",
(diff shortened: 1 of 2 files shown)


paulswartz:
Looks fine.

PR #264, opened 2012-03-09
Fix18851

bendk:
Couple changes to implement #18851.

### DIFF
--- a/tv/lib/frontends/widgets/firsttimedialog.py
+++ b/tv/lib/frontends/widgets/firsttimedialog.py
@@ -56,7 +56,7 @@ WIDTH = 475
 HEIGHT = 375
 
 
-def _build_title_question(text):
+def _build_title(text):
     """Builds and returns a title widget for the panes in the
     First Time Startup dialog.
     """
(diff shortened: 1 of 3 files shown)


tv/lib/frontends/widgets/firsttimedialog.py
```diff
@@ -524,50 +534,109 @@ def build_search_page(self):
...
-        lab.set_wrap(True)
-        vbox.pack_start(widgetutil.align_left(lab))
-
-        def handle_next(widget):
-            if rbg.get_selected() == yes_rb:
-                self.import_media_player_stuff = True
-            else:
-                self.import_media_player_stuff = False
-            self.next_page()
+        prefpanel.attach_boolean(vbox.net_lookup_cbx,
```
paulswartz: Can you use this to attach the lookup checkbox in the `MusicSetupVBox`?  Then you wouldn't need to look at the checkbox in the Music tab popup.

bendk: I was thinking about that, but then thought it was better not to because we can't handle the import checkbox inside MusicSetupVBox.  It seemed like if one gets handled outside the class, both should.

paulswartz: Why can't we move the `attach_boolean` call into the MusicSetupVBox?

bendk: We could, what we can't do is add the code to handle the "import from itunes" checkbox.  I thought it would be strange if we handled one checkbox inside MusicSetupVBox and one checkbox outside of it.

On the other hand, it would reduce code duplication.  I'll change it.

paulswartz:
Looks good.

PR #265, opened 2012-03-12
bz18780. force the label on ImageMenuItem (GTK)

paulswartz:
We were just using the stock label, which wasn't translated and didn't always
have the label we wanted.  Now we force the label after we create the menu
item.


bendk:
I think this is probably good, but I'm not 100% sure.

This seems like a question of if we want to use our translations or GNOME's translations for those menu items.  On windows, I think we definitely want to use ours, but do we want to do that on linux?

paulswartz:
Since we're using the 'Edit' image for the 'Translate' menu item, that shows up as 'Edit', which is confusing.

bendk:
Ahh, that makes sense.  Merging.

PR #267, opened 2012-03-12
Fix17857

bendk:


tv/lib/frontends/widgets/gtk/customcontrols.py
```diff
@@ -159,32 +154,78 @@ def gtk_scale_class(self):
...
             return
+        current_pos = self.get_slider_pos()
+        event_pos = self._event_pos(event)
+        pos_difference = abs(current_pos - event_pos)
+        # only move the slider if the click was outside its boundaries
+        # (#18840)
+        if pos_difference > self.slider_size() / 2.0:
+            self.move_slider(event_pos)
+        self.drag_info = _DragInfo(event.button,
+                                   self.get_slider_pos(),
```
paulswartz: Can be `current_pos`?

tv/lib/frontends/widgets/gtk/customcontrols.py
```diff
@@ -376,14 +429,38 @@ def set_value(self, value):
...
+        value
+        """
+        return self._widget.get_slider_pos(value)
+
     def set_range(self, min_value, max_value):
         self._widget.set_range(min_value, max_value)
         # Try to pick a reasonable default for the digits
         range = max_value - min_value
-        self._widget.set_digits(int(round(math.log10(100.0 / range))))
+        self._widget.set_digits(int(round(math.log10(10000.0 / range))))
```
paulswartz, line 453: What does this magic # do?

tv/osx/plat/frontends/widgets/customcontrol.py
```diff
@@ -198,6 +200,9 @@ def sendAction_to_(self, action, to):
         return YES
 DragableDrawableButton.setCellClass_(DragableButtonCell)
 
+MouseTrackingInfo = collections.namedtuple("MouseTrackingInfo",
```
paulswartz, line 203: Is there a reason this is implemented differently from `_DragInfo` on GTK?

tv/lib/frontends/widgets/gtk/customcontrols.py
bendk: current_pos is the position at the start of the event.  If we move the slider, we need to call get_slider_pos() again.

tv/osx/plat/frontends/widgets/customcontrol.py
```diff
@@ -251,6 +294,25 @@ def isOpaque(self):
...
+        if wrapper.is_horizontal():
+            size = self.bounds().size.width
+        else:
+            size = self.bounds().size.height
+        size -= slider_size
+
+        range = self.maxValue() - self.minValue()
+        value_change = (event.deltaY() / size) * range
+        self.setFloatValue_(self.floatValue() + value_change)
+        wrapper.emit('changed', self.floatValue())
```
paulswartz, line 315: Why don't we emit the 'pressed' and 'released' signals here?

tv/osx/plat/frontends/widgets/customcontrol.py
```diff
@@ -352,7 +417,7 @@ def set_range(self, min_value, max_value):
         self.view.setMinValue_(min_value)
         self.view.setMaxValue_(max_value)
 
-    def set_increments(self, increment, big_increment):
+    def set_increments(self, small_step, big_step, scroll_step=None):
         pass
```
paulswartz, line 428: The documentation from line 301 should probably be here as well.

tv/lib/frontends/widgets/gtk/customcontrols.py
bendk, line 453: I'm not 100% sure, but I think it sets the precision of the slider based on the number of digits.  Before we would use 2 digits of precision, but that made using the scrollwheel weird, since we could be seeking by less than 0.1.  After the change we get 4 digits of precision.

tv/osx/plat/frontends/widgets/customcontrol.py
bendk, line 203: I think there might be slight differences across platforms, but mostly I just didn't take the time figure out how to put the code in portable.

tv/lib/frontends/widgets/gtk/customcontrols.py
paulswartz, line 453: Can you put that in a comment?

tv/osx/plat/frontends/widgets/customcontrol.py
bendk, line 315: oops, I'll add that.

tv/osx/plat/frontends/widgets/customcontrol.py
bendk, line 428: I thought that we decided that GTK was the official place for docstrings for platform-specific code.  The reason not to put it in both is then when we change things, we need to update it in 2 places and probably won't.

paulswartz, line 428: I mean this:

``` python
# NOTE: we ignore the scroll_step value passed into set_increments()
# and calculate the change using deltaY, which is in device
# coordinates.
```

It is platform-specific.  It doesn't need to be a docstring; just a note about why we ignore the values.

bendk, line 428: Oops, totally misread that one.  Fixed now.

tv/lib/frontends/widgets/gtk/customcontrols.py
paulswartz: On OS X, that's done by resetting `current_pos = self.get_slider_pos()` and using the variable.  That avoids calling `get_slider_pos()` twice in some of the cases.

PR #270, opened 2012-03-13
bz18863: Don't create display until we really need to show it.

geoffl:
We were pre-creating the display and then passing the display to the
delayed timeout callback, but we should create the display in the delayed
timeout callback since it is only at that point when it is needed.  Saves
unnecessarily creating/freeing display when navigating the sidebar.


bendk:
Looks great to me.

PR #272, opened 2012-03-14
bz18607. update search icons

paulswartz:
This also adds a script which will grab the icons and make them PNGs
automatically.


geoffl:
looks good

PR #273, opened 2012-03-14
Just log when item info cache fetch fails.

geoffl:
Don't know what I was thinking at the time but failed() definitely can kick
up a crash report too so let's just run logging.info() and be done with it.

A crash here is really bogus since it is (a) just a cache and (b) we know
there are conditions where this can happen.


bendk:
Seems good to me.

PR #276, opened 2012-03-15
bz18870: disconnect TableView from TableModel when we're done with them.

bendk:
Before this was happening:
- User selects a tab, we create a table model and table view connected to
  that model.
- User moves the tab into/out of the folder
  - We re-use the table model and make a new table view
  - The old table view still would still handle the signals from the model,
    but we weren't really prepared to handle them and would throw an
    exception.

Fixed this by adding code to explicitly make a table view stop using a model.
It would be nice to support set_model() and make setting it to None a special
case, but it's not easy to do on OSX, so let's wait for this.

### DIFF
--- a/tv/osx/plat/frontends/widgets/tableview.py
+++ b/tv/osx/plat/frontends/widgets/tableview.py
@@ -1283,14 +1283,21 @@ class TableView(CocoaSelectionOwnerMixin, CocoaScrollbarOwnerMixin, Widget):
         self.header_height = HEADER_HEIGHT
         self.set_show_headers(True)
         self.notifications = NotificationForwarder.create(self.tableview)
-        self.model.connect_weak('row-changed', self.on_row_change)
-        self.model.connect_weak('structure-will-change',
-                self.on_model_structure_change)
+        self.model_signal_ids = [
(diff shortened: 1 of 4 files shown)


tv/osx/plat/frontends/widgets/tableview.py
```diff
@@ -1283,15 +1283,22 @@ def __init__(self, model, custom_headers=False):
         self.header_height = HEADER_HEIGHT
         self.set_show_headers(True)
         self.notifications = NotificationForwarder.create(self.tableview)
-        self.model.connect_weak('row-changed', self.on_row_change)
-        self.model.connect_weak('structure-will-change',
-                self.on_model_structure_change)
+        self.model_signal_ids = [
```
paulswartz, line 1286: Earlier in `__init__`, there's this line:

``` python
self.data_source.initWithModel_(self.model)
```

Should we be un-initing the data source as well?

bendk, line 1286: Good point.  I think the best way to clear this one out is to un-set the data source for the NSTableView.  I'll add a changeset for this one.

paulswartz:
I think your changes look good, and 17823 is still fixed with your new code.  Merging.

PR #277, opened 2012-03-16
bz18871.  in do_activate NameError: global name 'sender' is not defined

paulswartz:
In c2fe52ef the code got refactored a bit, and the code in question got moved
into a different function.  Since we're handling the signal before it goes out
to other handlers, what used to be the `sender` is now `self`, since the
current object is sending the signal.


bendk:
Looks great.

PR #278, opened 2012-03-19
bz18874. stock buttons not translated (GTK)

paulswartz:
For the same reason as #18780, we weren't translating stock buttons on GTK.
Now, we keep track of the text to use and set it manually.

I looked into including the GTK translations, but that's tricky on Windows
where we basically don't care about what the system preference is.  On Linux,
if you've set the system language to something else, this continues to DTRT.


bendk:
Looks good to me.  Let's merge before the RC.

PR #279, opened 2012-03-20
bz18883: remove progress bar when importing metadata.

bendk:
Refactored things slightly now that we only have 2 modes of display, but
overall I tried not to change too much.

### DIFF
--- a/tv/lib/frontends/widgets/itemlistcontroller.py
+++ b/tv/lib/frontends/widgets/itemlistcontroller.py
@@ -135,7 +135,7 @@ class ProgressTrackingListMixin(object):
             meter.update(self.mediatype, finished, finished_local, eta, total)
             self.postponed = None
         else: # got progress before widget created
-            self.postponed = (remaining, eta, total)
+            self.postponed = (finished, finished_local, eta, total)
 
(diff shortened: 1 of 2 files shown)


paulswartz:
Kind of related to this; there's a pyflakes error in tv/lib/frontends/widgets/itemlistcontroller.py:138

``` python
    def update_metadata_progress(self, finished, finished_local, eta, total):
        meter = self.widget.get_progress_meter()
        if meter:
            meter.update(self.mediatype, finished, finished_local, eta, total)
            self.postponed = None
        else: # got progress before widget created
            self.postponed = (remaining, eta, total)
```

`self.postponed` should be set to the same arguments as the update call, possibly like this?

``` python
    def update_metadata_progress(self, finished, finished_local, eta, total):
        meter = self.widget.get_progress_meter()
        args = (self.mediatype, finished, finished_local, eta, total)
        if meter:
            meter.update(*args)
            self.postponed = None
        else: # got progress before widget created
            self.postponed = args
```

bendk:
Good catch on the that one.  I just added a changeset to address it.

geoffl:
looks good

PR #280, opened 2012-03-20
bz18881: fix attribute error for value()

bendk:
Not all table models have iters that define value().  The correct way to get
the value of the row for all models is to use index notation.


paulswartz:
Not my review day, but looks good to me.

geoffl:
let's merge

PR #281, opened 2012-03-21
bz18885. catch OSError when trying to get the name of the device

paulswartz:
We were raising OSError when getting the name, but only catching WindowsError.
There's probably another bug in here somewhere, but without being able to get
more information from the reporter (it came in through Bogon) all I can do is
not have it crash Miro.


geoffl:
looks good

PR #282, opened 2012-03-21
bz18882: fix unittests when mutagen returns a slightly different duratio...

bendk:
...n.
- Use assertDictEquals in filetagstest.  This outputs better messages if the
  comparison fails.
- Added a assertClose() method to test if 2 values are near each other.
  There's a assertNearlyEqual method from UnitTest, but that one didn't do
  what we wanted.
- Use assertClose() to compare durations from mutagen.
- While we're at it, use assertClose() to test preallocation code.


bendk:
Wikipedia suggests that one, but using the max value:

http://en.wikipedia.org/wiki/Relative_change_and_difference

I'm going to go with that.

geoffl:
looks good

PR #283, opened 2012-03-21
bz18888: Add set_size() to Checkbox on GTK

bendk:
Made Checkbox create its own Label and use that rather than relying on the one
that GTK creates by default.

While we're at it, we can now support the bold parameter.


geoffl:
looks good

PR #284, opened 2012-03-21
bz18886. clear icon cache when changing items

paulswartz:
YouTube video names are pretty common (3.*.jpg), so when we weren't clearing
the cache when changing the underlying thumbnail, the frontend was getting
confused.

### DIFF
--- a/tv/lib/feed.py
+++ b/tv/lib/feed.py
@@ -57,13 +57,12 @@ from miro import dialogs
 from miro import download_utils
 from miro import eventloop
 from miro import feedupdate
-from miro import flashscraper
 from miro import models
 from miro import prefs
 from miro.plat import resources
 from miro import downloader
 from miro.util import (returns_unicode, returns_filename, unicodify, check_u,
-                       check_f, quote_unicode_url, escape, to_uni,
+                       check_f, quote_unicode_url, to_uni,
                        is_url, stringify, is_magnet_uri)
 from miro import fileutil
 from miro.plat.utils import filename_to_unicode, make_url_safe, unmake_url_safe
@@ -101,7 +100,6 @@ class _RateLimiter(object):
         self.last_time = time.time()
 
     def check_for_sleep(self):
-        new_time = time.time()
         elapsed = self.last_time - time.time()
         if elapsed < 0.1:
             return # don't sleep until a decent of time has passed.
@@ -239,7 +237,7 @@ def run_feedparser(html, callback, errback):
     else:
         workerprocess.send(workerprocess.FeedparserTask(html),
                            lambda msg, result: callback(result),
-                           lambda msg, error: errback(result))
+                           lambda msg, error: errback(error))
 
 # Wait X seconds before updating the feeds at startup
 INITIAL_FEED_UPDATE_DELAY = 5.0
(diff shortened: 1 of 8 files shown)


bendk:
Seems okay, but it seems like the issue is more that the filename is always the same, rather than it's just short.  Maybe we should target the 3.jpg specifically, rather than short filenames?

Maybe this is a bit of a stretch, but i'm thinking of a situation where the URL path is always the same, but the server picks the file and sends a content-disposition header based on the query.  In that case, we will now be always using the same filename and could run into the same error.

paulswartz:
Yes, that's true; this would happen for anything where the filename doesn't change.  I'll think about a better way to do this.

paulswartz:
Okay, reopened for real.  The issue is actually different, but what I did fixed it accidentally.  The real issue was that we weren't clearing the frontend image cache, so we'd sometimes see the old image because it wouldn't be reloaded from the disk.

tv/lib/iconcache.py
```diff
@@ -217,6 +217,11 @@ def update_icon_cache(self, url, info):
                 return
 
             filename = unicode(info["filename"])
+            if len(filename) < 8:
+                # bz18886. YouTube thumbnail URLs are easily broken by this;
+                # the filenames are all 3.jpg.  Lets use a version of the URL
+                # instead.
+                filename = unicode(url.split('://', 1)[1][-50:])
```
bendk: I still think we should only target when filename == '3.jpg' or whatever the youtube one is.  Seems like shortfile name should be fine as long as their unique and the only one we know of that isn't is the youtube one.

tv/lib/frontends/widgets/application.py
```diff
@@ -1570,6 +1571,14 @@ def handle_item_list(self, message):
...
 
     def handle_items_changed(self, message):
+        if message.changed:
+            # all thumbnails have this in the name; the / in the path means it
+            # can't be part of any generated name
+            default_thumb = os.path.join('images', 'thumb-default')
+            for item in message.changed:
+                thumbnail = item.thumbnail
+                if default_thumb not in thumbnail:
+                    imagepool.clear_cache_for_path(thumbnail)
```
bendk: should we have this code in handle_item_list() as well?  It seems like its possible for this to happen there too.  However, I'd be worried about performance if we did it for every new item.  Maybe we should just stick with this, then add a note to investigate more after 5.0?

paulswartz: Technically (and I should have put a note in about this), we should be doing it on remove instead of change.  But since `message.removed` just has the ID #s, changed was the best thing I can think of.  New items should be fine; they shouldn't be in the cache and that's why I'm not checking `message.added` in this method.

bendk: So the issue is that when we remove the items, we keep the image object in the imagepool cache, keyed with the filename and if we get that filename again, we use the old image right?

I think we could still run into the issue with the items in message.added then, it's just that we don't because of the timing of iconcache updates.

Could we add an mtime to the imagepool cache and check that when we want to fetch images?

paulswartz: Yeah, I can do that.

paulswartz:
Just pushed up a new changeset which implements the mtime check.

bendk:
Looks great to me.

PR #289, opened 2012-03-22
bz18893: Don't crash on KeyErrors in devices.py:on_new_metadata

bendk:
I think can happen under normal, if slightly unlikely, conditions if you get
metadata updates, but disconnect the device before the update timeout happens.
Seem like a warning is appropriate.


paulswartz:
Looks good.

PR #290, opened 2012-03-23
bz18897: avoid decoding errors when printing our path.

bendk:
I'm pretty sure we should always be using %r to format paths, since it works
for both unicode, bytestrings, and even if there's an invalid character in
there.


paulswartz:
Looks good.

PR #291, opened 2012-03-23
Fix18890

bendk:
Fixing this bug required a going on a couple side  journeys and a fairly complex couple of changesets.


paulswartz:
Merging.

PR #292, opened 2012-03-26
bz18895. don't create .miro on hidden devices

paulswartz:


bendk:
I think this is okay, but it seems a bit un-clean to modify the path attribute like that.  Seems like the LiveStorage constructor could get an argument that makes it start in temp mode, and not try to switch out of it.

However, that solution is more complicated.  If this has been well tested, then I'm okay doing it for the release, and making a TODO to clean things up better.

What do you think paul?

paulswartz:
No, I think that's a better fix.  I'll make that change today.

paulswartz:
Implemented a `start_in_temp_mode` for LiveStorage.

bendk:
The start_in_temp_mode change looks correct, but are we getting the same behavior afterwards?  Will it try to save the directory when the temp mode timeout stops?

After I wrote that I'm wondering if it still does the right thing because the .miro directory is not in place yet, so the attempt fails.  If so, that seems fine but a comment would be good.

paulswartz:
It doesn't create the .miro directory at first because we're setting `force_directory_creation` to `False` before the directory is created.  Once the device is shown, we create the directory and switch out of temporary mode.

bendk:
Looks great.  Merging.

PR #295, opened 2012-03-26
Fix18894

bendk:
I wasn't sure exactly how this bug happened, but hopefully this is good enough for a fix.  One changeset turns the crash into a soft failure, since I don't think it's the biggest deal if it happens.  The other changeset fixes some of pieces of code that could be the cause.


paulswartz:
Looks fine.

PR #296, opened 2012-03-26
bz18915: fix crash in GTK custom slider code

bendk:
self.drag_info will be None if the user has 2 buttons pressed down at once.
Then when the first button release happens we set set drag_info to None and
emit our signals.  When the second button is released we should just ignore
it.


paulswartz:
Seems harmless enough.

PR #297, opened 2012-03-28
bz18918: Use EnvironmentError to catch shutil.Error.

geoffl:
Catch hopefully all possible errors when trying shutil.copyfile() fails.


bendk:
This is definitely better.

PR #298, opened 2012-03-28
bz18917: Some defensive programming for self.playbin.

geoffl:
Most likely in the nest of callbacks and signaling the playbin was destroyed.
I don't want to spend too much time to figure out why; the reason is probably
mostly harmless.  Put in some checks and all should go well.


paulswartz:
Looks fine to me.

PR #299, opened 2012-03-28
bz18919: Update 5min search url.

geoffl:
It has changed; update it.


bendk:
Trival change, looks great to me.

PR #300, opened 2012-03-28
bz18930: Guard against get_total_playback_time() use.

geoffl:
This method may return None, so guard against its usage in arithmetic
operations.  In addition it may theoretically return 0.  Make sure
we never use such a value in the denominator.

### DIFF
--- a/tv/lib/frontends/widgets/playback.py
+++ b/tv/lib/frontends/widgets/playback.py
@@ -270,7 +270,11 @@ class PlaybackManager (signals.SignalEmitter):
         if self.player_playing():
             elapsed = self.player.get_elapsed_playback_time()
             total = self.player.get_total_playback_time()
-            self.emit('playback-did-progress', elapsed, total)
+            if elapsed is not None and total is not None:
+                self.emit('playback-did-progress', elapsed, total)
+            else:
+                logging.warning('notify_update: elapsed = %s total = %s',
+                                elapsed, total)
 
     def on_display_removed(self, display):
         if not self.removing_video_display:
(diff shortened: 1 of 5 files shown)


tv/lib/frontends/widgets/playback.py
```diff
@@ -454,6 +458,9 @@ def update_current_resume_time(self, resume_time=-1):
         if resume_time == -1:
             resume_time = self.player.get_elapsed_playback_time()
             duration = self.player.get_total_playback_time()
+            if duration is None:
+                logging.warning('update_current_resume_time: duratio is None')
```
paulswartz: typo

geoffl:
ok so I think I basically agree with all the feedback here.  The duration == 0 case was checked as an afterthought so you are right some of the tests are not quite right.

tv/lib/frontends/widgets/playback.py
```diff
@@ -282,6 +286,12 @@ def play(self, start_at=0):
             logging.warn("no self.player in play(). race condition?")
             return
         duration = self.player.get_total_playback_time()
+        if duration is None:
+            logging.warning('duration is None.  race condition?')
+            return
+        # let this one slide but print out big fat warning
+        if duration <= 0:
+            logging.warning('duration is %s', duration)
         self.emit('will-play', duration)
```
paulswartz: Should we be emitting the duration if it's <= 0?

geoffl: Good question!  Thought I'd let it slide and print out a big warning.  I guess 0 is okay but maybe not < 0.  Still, there could be a bug with the decoder and the duration cannot be returned and yet the file is still playable?

paulswartz: It looks like the code that actually uses the `duration` in this call is all in https://github.com/pculture/miro/blob/master/tv/lib/frontends/widgets/videobox.py, in the `handle_play()` methods.  At least one of these, `ProgressSlider.handle_play`, will break if given 0: https://github.com/pculture/miro/blob/master/tv/lib/frontends/widgets/videobox.py#L343

geoffl: I was trying to look for places where this might happen and it looks like I might have missed this particular case.  I think that was the only one.  But I think it's more convincing to just return if <= 0 rather than continuing.

paulswartz:
Looks good now.

PR #301, opened 2012-03-28
bz18925: Change Metavid search url.

geoffl:
The url's changed; change the corresponding search engine entry.


bendk:
Trivial change.

PR #302, opened 2012-03-28
bz18902: Make sure that embedding window is still valid on destructor pa...

geoffl:
...th.

Make sure the embedding window is still around when calling destroy
as the destructor path may stil make use of it.

### DIFF
--- a/tv/windows/plat/frontends/widgets/embeddingwidget.py
+++ b/tv/windows/plat/frontends/widgets/embeddingwidget.py
@@ -98,10 +98,15 @@ class EmbeddingWidget(gtk.DrawingArea):
             self.embedding_window.reposition(*self._get_window_area())
 
     def destroy(self):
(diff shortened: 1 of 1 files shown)


paulswartz:
How does the GTK destructor know about the embedding window?  Also, this should change/remove the comment so that it matches the code.

geoffl:
The GTK destroy() does not but it calls do_unrealize() apparently on the destroy() path which can make use of the embedding widget.

bendk:
The original reason to put our destroy call before the GTK destroy call was because our window is a child of the GTK window.  However, if destroy() makes sure to call unrealize() if needed, then we should be okay, since we will switch parents at that point.

So I think the code is okay, but can you a) double-check my logic there and b) add a comment describing what's happening.

paulswartz:
Looks like bendk's review comments have been addressed.

PR #303, opened 2012-03-28
bz18929. don't convert items we can't handle

paulswartz:
`conversion_for_info()` returns None for items it can't handle (they don't have
a path, they aren't media), and we shouldn't try to do anything with them.


geoffl:
looks good

PR #304, opened 2012-03-28
bz18927. `title_tag` no longer part of metadata

paulswartz:
We don't have it as an attribute, so don't try to copy it.


geoffl:
looks good

PR #305, opened 2012-03-28
don't break the unittests

paulswartz:
The unittests don't go through the regular code path here, and I missed that.  My fault.


geoffl:
looks good let's merge

PR #307, opened 2012-03-28
bz18926. cancel/ignore in-progress metadata updates on device eject

paulswartz:
We weren't cancelling them, so when the results came back, we'd try to write to
the now-closed database.


bendk:
Looks good.

PR #308, opened 2012-03-28
Fix18806

bendk:
Bunch of fixes for #18806.  I'd love review, but please don't merge until Janet says that its fast enough on her machine.

### DIFF
--- a/tv/lib/metadata.py
+++ b/tv/lib/metadata.py
@@ -599,7 +599,7 @@ class _EchonestProcessor(_MetadataProcessor):
         if not self.should_skip_codegen(current_metadata):
             self._codegen_queue.add(path)
         else:
-            self._echonest_queue.add((path, None))
+            self._echonest_queue.add(path, None)
         self._process_queue()
 
     def should_skip_codegen(self, metadata):
--- a/tv/lib/metadata.py
+++ b/tv/lib/metadata.py
@@ -173,6 +173,12 @@ class MetadataStatus(database.DDBObject):
         return [r[0] for r in rows]
 
     @classmethod
+    def net_lookup_enabled_view(cls, net_lookup_enabled, db_info=None):
+        return cls.make_view('net_lookup_enabled=?',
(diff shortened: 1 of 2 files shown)


tv/lib/metadata.py
```diff
@@ -698,6 +713,38 @@ def _should_process_codegen_queue(self):
...
+                                      self._restart_after_http_errors, name)
+                self._waiting_from_http_errors = True
+
+    def _process_metadata_fetch_queue(self):
+        while self._metadata_fetch_queue:
+            path, metadata_fetcher = self._metadata_fetch_queue.pop()
+            try:
+                metadata = metadata_fetcher()
+            except KeyError:
+                logging.warn("_process_metadata_fetch_queue: metadata_fetcher "
```
paulswartz, line 739: Is this log item going to be useful? Some things I can think to include:
- item
- fetcher function
- traceback

Also, what about other exceptions?

bendk, line 739: Yeah, all of those things would be good.  By item, I'm going to use path.

I was thinking that other exceptions should be propagated upward, but now I don't think so anymore.  I'll change to catch StandardError.

bendk:
Also, janet checked out the branch and it was running fast enough for her.  So this can be merged once it passes review.

tv/lib/metadata.py
```diff
@@ -1471,23 +1527,25 @@ def _run_movie_data(self, path):
...
-        # we only send a subset of the metadata to echonest and some of the
-        # key names are different
-        echonest_metadata = {}
-        for key in ('title', 'artist', 'album', 'duration'):
-            try:
-                echonest_metadata[key] = metadata[key]
-            except KeyError:
-                pass
-        self.echonest_processor.add_path(path, echonest_metadata)
+        def metadata_fetcher():
```
paulswartz, line 1532: Do we still need this comment, since we're doing it later to avoid the slowdown?

paulswartz:
Looks good!

PR #310, opened 2012-03-30
bz18914. don't show read-only devices

paulswartz:
If we can't write to the root of the device, don't bother creating a database
or even showing it to the user.  This prevents lock errors on SQLite (can't
lock something on a read-only device on Windows), and saves CPU since we won't
scan the device.


geoffl:
looks okay

PR #311, opened 2012-04-02
bz18957. avoid overflow in pango.PIXELS

paulswartz:
A user experienced a crash where the result of ascent() + descent() wouldn't
fit in an int.  I can't reproduce it, but this should avoid the issue.


geoffl:
looks good

PR #314, opened 2012-04-04
bz18965. OSError in mtime_invalidator

paulswartz:
If the file does not exist, or disappears, the cache should be invalidated.


geoffl:
looks good

PR #316, opened 2012-04-04
bz18974: Always emit will-play.

geoffl:
Many components depend on it so emit even if the duration is None.
Having said that make the receivers of this signal check that duration
is vaild before using.


bendk:
Looks good.

PR #317, opened 2012-04-04
bz18972: Update appcast urls.

geoffl:


bendk:
Pretty simple, looks good to me.

PR #319, opened 2012-04-05
bz18644, comment #18: increase min-width of the date column.

bendk:


paulswartz:
Just increases the width of the column; looks fine.

PR #323, opened 2012-04-09
bz18969. fallback for Natty

paulswartz:
Natty has an older version of FFmpeg, so instead of the new commandline
arguments we need to use older versions.


bendk:
Looks fine to me.

PR #324, opened 2012-04-09
bz18992: fix layout of dialogs on OSX (specifically item-edit)

bendk:
Added a function to help debug layout code.  It recursively prints out the
size request for a given widget and its descendents.

Fixed things 2 ways:

 1) If the size request for our content widget changes, then we re-place it.
 This is needed because the window frame size changes.

 2) Set size request for the item edit content panel so that it fits the
 biggest panel we have.  This makes the dialog change less when things switch.


paulswartz:
Looks fine.

PR #329, opened 2012-04-12
Updated setup.py and binary_kit_version

bendk:
This goes with a binary kit change that adds the DLLs from Visual C++ 10.0.


paulswartz:
Looks fine.

PR #330, opened 2012-04-12
bz19008. timeout MutagenTask (if supported)

paulswartz:
When given a broken file, Mutagen can hang in a busyloop reading the file.  If
supported by the operating system, we can use signals (SIGALRM in particular)
to kill the busyloop.

I'm not convinced that this is the ideal way to do it, but that'll get covered
by peer review.


bendk:
Looks great.  Merging.

PR #333, opened 2012-04-19
bz19033: handle items being removed in-between idle-iterator steps

bendk:
Fetch all items at once and store them in a list rather than iterate through
self.items.  We shouldn't be using the same view after we yield and let other
threads access the DB.

Check for items being removed in-between when we've noticed their dups and
when we get around to deleting them.

These are both blind fixes, but I think they should be safe.  I'm also fairly
positive they will fix the bug.


paulswartz:
Looks fine.

PR #334, opened 2012-04-20
Revert "bz18863: Don't create display until we really need to show it."

paulswartz:
This reverts commit d8bc1d5b37959f80b2f503d2bafe11842376e7d9.  It's an efficiency
speedup, but it's causing the widths of views to be calculated incorrectly.


bendk:
Looks good to me.  Merging.  Can you re-open #18863 paul and add comments explaining whatever you know about it.

PR #335, opened 2012-04-20
bz18985. fix failing MovieData tests on OS X

paulswartz:
DRM files were handled slightly differently on OS X, so we special-case them in the unit-tests.


bendk:
unittest-only changes, seems fine to me.

PR #338, opened 2012-04-23
bz19040. some feeds break feedparser's image handling

paulswartz:
I didn't actually see this with my version of Feedparser (it's returning a
dictionary with an 'href' key instead), but I can visualize it returning a
Unicode string given some feed.  This code tries to be a bit more careful about
checking the keys (and whether it's looking at a dictionary at all) before
overriding the thumbnail.


bendk:
Looks great.

PR #339, opened 2012-04-23
bz19071: better debugging to figure out this problem.

bendk:
This change definitely won't fix the bug, but it does help improve logging in
2 ways:
- Added extra info when we call failed_soft()
- Fix for stringify(), which is called by crashreporter so that it doesn't
  escape newlines.  Otherwise reading the log reports is pretty miserable.

I've tested this code by provoking the crash and it seems okay.


paulswartz:
Ahhh, thank you for the stringify change! Looks good.

PR #340, opened 2012-04-23
Fixed stringify() and its unittest.

bendk:
My recent change to stringify() made it return unicode objects, which is
definitely wrong.  Added a call to encode afterwards to fix this.  Use the
replace error handler both ways since only that and ignore work for both
decoding and encoding.

Fixed the unittest so it passes.


paulswartz:
Looks fine.

PR #341, opened 2012-04-24
bz19017: Dispose of our thumbnail pipeline to avoid memory leaks

bendk:
Before when I put 200 or so videos through movie data miro_helper.exe would
use over 500 megs of virtual memory.  Now it's around 70-80megs.

This actually only fixes part of #19017, but I think it's an important one.


paulswartz:
Looks good to me.

PR #342, opened 2012-04-24
bz19032: better debugging for failed SQL statements.

bendk:
Log errors when our SQL statements fail for errors other than OperationError.
This includes integrity error and others.

Refactored log_error slightly

This won't fix #19032, but hopefully it will help us track it down.


paulswartz:
Looks fine.

PR #343, opened 2012-04-27
bz19032: metadata fixes for items marked as deleted

bendk:
If an item is marked as deleted then we delete the metadata entries for it,
however there were 2 bugs with that.

1) The next time we called setup_restore(), we would re-create the metadata
entries.

2) For items deleted with previous versions of miro, our upgrade code would
make metadata entries for it.

Fixed those 2 bugs and also added code to detect this problem and log a
warning instead of crashing.


bendk:
This is super late in the release process, so please give try to review this one extra good.

paulswartz:
I think it looks okay, but I'm going to hold off on merging until 1) Geoffrey looks at it 2) tomorrow.

geoffl:
I think it looks okay too, maybe we can just test it extra well.

paulswartz:
Merging.

PR #344, opened 2012-04-28
fix failing MovieDataTest on Jenkins

paulswartz:
Some builds of FFmpeg can't handle the WebM format video we're testing with.
I've moved that test case over to the OS X side, and testing with the Theora
video file instead.


bendk:
Seems good to me.  And it only changes the unittest code, so I think it's safe to merge at this point.

PR #351, opened 2012-05-11
bz19099: switch to write-ahead logging mode

bendk:
This should make things generally faster, and especially help us when we have
the frontend reading from and maybe even writing to the DB.

### DIFF
--- a/tv/lib/storedatabase.py
+++ b/tv/lib/storedatabase.py
@@ -455,15 +455,31 @@ class LiveStorage:
                     raise
 
         self.cursor = self.connection.cursor()
+        if path != ':memory:' and not self.temp_mode:
+            self._switch_to_wal_mode()
+
+    def _switch_to_wal_mode(self):
+        """Switch to write-ahead logging mode for our connection
+
+        WAL mode allows for better concurency between readers and writers and
+        is generally faster than other modes.  See:
+        http://www.sqlite.org/wal.html
+        """
         try:
-            self.cursor.execute("PRAGMA journal_mode=PERSIST");
+            self.cursor.execute("PRAGMA journal_mode=wal");
         except sqlite3.DatabaseError:
-            msg = "Error running 'PRAGMA journal_mode=PERSIST'"
+            msg = "Error running 'PRAGMA journal_mode=wal'"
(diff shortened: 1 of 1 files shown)


tv/lib/storedatabase.py
```diff
@@ -455,15 +455,31 @@ def open_connection(self, path=None, start_in_temp_mode=False):
...
             self.created_new = True
             # rerun the command with our fresh database
-            self.cursor.execute("PRAGMA journal_mode=PERSIST");
+            self.cursor.execute("PRAGMA journal_mode=wal");
+        # check that we actually succesfully switch to wal mode
+        actual_mode = self.cursor.fetchall()[0][0]
+        if actual_mode != u'wal':
+            msg = ("PRAGMA journal_mode=wal didn't change the mode.  "
+                   "journal_mode=%s" % actual_mode)
+            self._handle_load_error(msg, init_schema=False)
```
paulswartz: Is it actually an error if SQLite doesn't switch? It'll be slower, sure, but shouldn't we continue if possible?

geoffl:
I don't think this will work on Windows for now due to an ancient sqlite all being shipped with Python?

geoffl:
also WAL does not work on network drives.  What happens when you do this to a sqlite database when mounted on a network drive?  Probably the load error message won't be very informative in this case.

bendk:
Okay, let's just log a warning if switching fails.  Geoff, can you give me a pointer on info about the sqlite issue on windows?

bendk:
Checked this out and indeed there is an issue with the sqlite version on winodws.  The newest versions from google code seem to work.

I'm going to make a ticket to track the windows binary stuff for version 6.0.  I'd like to try to use some of the code to make a virtualenv for mvc in miro.

bendk:
Sorry for the close, hit the button by accident.

paulswartz:
Ben, what's the status for this? Does the warning cover the Windows case?

bendk:
I think it can be merged.  On windows we don't support WAL mode yet, but that just means we'll get the warning.  I'll open a new bug to upgrade SQLite on windows.

PR #353, opened 2012-05-15
Fix19019

bendk:
While fixing this one I ended up finding a bunch of subtle errors and added changesets for those as well.

If you review this, can you try to be extra-careful?  I'd like to put this code into the 5.0.1 branch, and it's not the simplest thing for janet to try to test, so let's get strong peer-review for it.


paulswartz:
Merging.

PR #354, opened 2012-05-18
bz19066. more working around Vimeo flakiness

paulswartz:
Now, we use Vimeo's built-in (sort of) support for getting download URLs.  It also adds a test case for Vimeo scraping.


bendk:
I think this should be fine.  I had one small nitpick/question.  If you want to change something based on that, go for it.  If I'm not reading the regex right, ignore it.

Either way, I think it's safe to merge and you should go for it once you either make the change or decide to ignore.

PR #355, opened 2012-05-18
bz19136: Fix race condition in sending the MovieDataTaskStatus message

bendk:
We need to send MovieDataTaskStatus before we send the TaskResult.  Otherwise
there's a possibility that the main process will see the TaskResult, then the
timeout will fire, before MovieDataTaskStatus gets processed.  This all
results in the crash that janet reported.

I'm hoping to get this one into the 5.0 branch, so please give it a nice peer-review.


paulswartz:
Looks like you're just sending the status messages more closely to where you're handling the tasks, and sending the "we're done" message a little earlier.  Seems fine.

PR #358, opened 2012-05-21
Changes to the title column of the item table.

bendk:
Before title tracked the title that we got from metadata (file tags, echonest,
etc).  Now that's stored in metadata_title, and title is the computed value of
title (what get_title() used to return).

This is needed if we are going to implement frontend item list using SQL
queries, since if the user sorts by title, we need this for the ORDER BY
clause.

I wrote this while working on a private branch to implement the plan in FrontendDataOverhaul.  I was thinking that maybe it would be a good idea to commit changes like this to master that are self-contained DB schema changes.  I think it might make merging less painful when I'm done with everything.

I could see an argument to not checking this in until I was done with all my work for FrontendDataOverhaul.  If you want to make an argument for that, then I could wait on merging.

### DIFF
--- a/tv/lib/item.py
+++ b/tv/lib/item.py
@@ -998,6 +998,7 @@ class Item(DDBObject, iconcache.IconCacheOwnerMixin):
     def update_from_metadata(self, metadata_dict):
         """Update our attributes from a metadata dictionary."""
         # change the name of title to be "metadata_title"
+        metadata_dict = metadata_dict.copy()
         if 'title' in metadata_dict:
             metadata_dict['metadata_title'] = metadata_dict.pop('title')
         self._bulk_update_db_values(metadata_dict)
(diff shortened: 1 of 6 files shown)


tv/lib/item.py
```diff
@@ -995,7 +997,11 @@ def file_moved(self, new_filename):
 
     def update_from_metadata(self, metadata_dict):
         """Update our attributes from a metadata dictionary."""
+        # change the name of title to be "metadata_title"
+        if 'title' in metadata_dict:
+            metadata_dict['metadata_title'] = metadata_dict['title']
```
paulswartz: Should this be `.pop('title')`?

bendk: Good point.  Pushed a new commit for this.

PR #362, opened 2012-05-24
Make plugins work on linux.

bendk:
I'm not sure the ramifications of changing gtk.gdk.threads_init() to
gobject.threads_init(), but it seems correct in theory at least.  We don't
need gtk.gdk.threads_init() unless we access GTK/GDK from separate threads,
which we don't.  And if we did that, we would need a
threads_enter()/threads_leave() pair, which we're not using.

This fixes things for me.  Is it really this easy?

### DIFF
--- a/tv/linux/plat/frontends/widgets/application.py
+++ b/tv/linux/plat/frontends/widgets/application.py
@@ -143,7 +143,7 @@ class LinuxApplication(Application):
         gobject.set_application_name(app.config.get(prefs.SHORT_APP_NAME))
         os.environ["PULSE_PROP_media.role"] = "video"
 
-        gtk.gdk.threads_init()
+        gobject.threads_init()
         self._setup_webkit()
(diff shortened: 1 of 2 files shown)


geoffl:
does gstreamer use threading with gobject/gtk?  otherwise I think this is good to go

paulswartz:
Should we be doing this on Windows, too?

We're not planning to merge this for 5.0, right? I'm fine with throwing it into master since we'll find any issues soon enough.

paulswartz:
Looks like Debian unbroke my libav libraries, so I was finally able to test this.  It works for me too, so I'm going to merge.

bendk:
Already merged, but I guess I'll add some comments.  I don't believe gstreamer does anything with GTK/GDK.  It simply inputs an x window id and renders to that.

Using gobject.threads_init() is exactly what we do on windows.  That's partly why I think this should be safe.

PR #363, opened 2012-05-24
Bumping binary kit version.

bendk:


bendk:
This goes with a binary kit upgrade for OSX that makes libtag work on both i386 and x86_64.  Asking glee to check it so I think he should be the one to merge this.

bendk:
Okay, now this change includes both the libtag and ffmpeg changes.  Janet still needs to test ffmpeg on ppc, but I think this should be good to go.

PR #364, opened 2012-05-25
Don't run the ENMFP codegen on PPC architectures on OSX.

bendk:
Our binaries aren't built to do that, so let's just skip it.


geoffl:
I think this is good to go. (I went in to fix the uname check to work for both i386 and x86_64)

PR #365, opened 2012-05-28
bz19018. don't change file handlers when reinstalling

paulswartz:
We weren't disabling the existing sections, so the default behavior was
happening instead.


bendk:
Looks great to me.

PR #370, opened 2012-06-13
bz17009: Update binary kit, ditch ppc, add x86_64.

geoffl:
- Fixed 64-bit size field for data passing in case we ever decide to run
  a mixed mode application suite with helpers.
- Bump minimum required version to 10.6.
- Support for building on Xcode4/Lion.
- Updated objective C signatures for a few functions where it needs to be set.
- Nuked 32-only GetMediaLanguage().
- _app_command_line(): either run legacy or native mode.

notes: the signature is probably not done the best way but it seems to me it is the most clear this way.  also, now needs xcode 4.  because the paths are different the build is now incompatible with xcode 3.


bendk:
Looks good.  I had one small comment and also a question: have you checked what the difference is now that we don't run GetMediaLanguage()?  I don't really know what the old code did, but my one concern is loosing whatever functionality it provided.

geoffl:
seems GetMediaLanguage is now unsupported and there is no equivalent.  I've always been confused by the code, I'm not convinced it does the right thing.  Let's just get it out of the file, if the language hint in there is garbage there does not seem to be much we can do about it.

geoffl:
one more thing: currently it builds for both x86_64 and i386 but i386 takes precedence because all the perian plugins are i386 only.  You can enable the x86_64 build by switching the ordering Info.plist but you won't be able to play perian supported formats.

bendk:
Looks good.  Merging.

PR #373, opened 2012-06-18
bz19228: fix errors with the retry_time column.

bendk:
Fixed crash when we failed to unpickle status because I spelled
"StandardError" wrong.

Fixed the underlying issue which was that we needed to support unpickling
datetimes in the database upgrade.

Lastly, fix the column type for retry_time so that it actually works.


paulswartz:
Looks fine.

PR #375, opened 2012-06-18
bz19231: fix descriptions for migrated items.

bendk:
Before we were using "" when we didn't have any description data from the
mutagen or other metadata sources.  Change that to None, which is the
standard, and what we use for new items.

### DIFF
--- a/tv/lib/databaseupgrade.py
+++ b/tv/lib/databaseupgrade.py
@@ -3966,3 +3966,7 @@ def upgrade183(cursor):
 def upgrade184(cursor):
     """Drop the seen column from item."""
     remove_column(cursor, 'item', ['seen'])
+
+def upgrade185(cursor):
+    """Use NULL for empty item descriptions."""
+    cursor.execute("UPDATE item SET description=NULL WHERE description=''")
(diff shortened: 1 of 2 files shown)


tv/lib/databaseupgrade.py
```diff
@@ -3966,3 +3966,7 @@ def upgrade183(cursor):
 def upgrade184(cursor):
     """Drop the seen column from item."""
     remove_column(cursor, 'item', ['seen'])
+
+def upgrade185(cursor):
+    """Use NULL for empty item descriptions."""
+    cursor.execute("UPDATE item SET description=NULL WHERE description=''")
```
paulswartz, line 3972: Is there a way we can tell the difference between an actual blank description (no description from the feed) and something that wasn't migrated correctly?  Is that a difference that even matters?

bendk, line 3972: So this is actually kind of weird because the feed description is stored in entry_description.  This is for the "metadata description", which i'm not sure if we even support at all right now.  As far as I can tell, it's always blank for all items.  I thought of just removing that column, but I could see us using it in the future if we could get descriptions from ID3 tags or something like that.

The only way that I could see to get a blank description is if you opened the item edit dialog and put nothing in there.  That might be a separate bug actually, since I think we should use NULL for that as well.

paulswartz, line 3972: If someone actually blanks out their description, I think we should respect that, rather than using something else.  But, I don't think that's a big enough deal for now.  Merging.

PR #376, opened 2012-06-20
Frontenddata

bendk:
I'm pretty sure this isn't ready to merge yet, but I'm making a pull request so that we can start peer-reviewing it.


bendk:
Geoff says this is okay to merge and janet has tested it a bit.  I'm going to go for it.

PR #378, opened 2012-06-26
bz19249: Catch errors that can be thrown from locale.getlocale().

geoffl:
... to prevent a crash on startup on malformed/invalid locale
settings by user.


bendk:
Looks fine to me.

PR #379, opened 2012-06-27
Blind fix for bz19227.

bendk:
Check if the metadata manager for a device is closed before making items on
it.


bendk:
I'm pretty sure we check if we should be stopping inside the idle iterator each time by testing the output of _stop().  The issue was that at the end of it all we were calling migrate_file() which uses a callback to indicate where it's done.

However, your comment made me realize that there was already a function to test what we wanted in a better way.  I added a commit to do that instead.

geoffl:
OK let's merge

PR #380, opened 2012-06-27
bz19273: convert text to utf-8 before sending it to gtk.Label

bendk:
I'm not really sure why it was crashing before, but this seems like a
safe change in any case.


geoffl:
looks good

PR #381, opened 2012-07-01
bz19323: Ditch ffmpeg2thora for ogv conversions.

geoffl:
ffmpeg binary can do all.


bendk:
Has this always been true, or did this change with an ffmpeg release?  If so, we should double check that this will work most recent linux distros.

geoffl:
I think it's always been true, even on debian stable with the crippled ffmpeg it is supported.  Not on Mac but I put the change into the binary kit that does it.

PR #383, opened 2012-07-05
One more pass ot #19227.

bendk:
Janet found a way to have the metadata manager be closed, but all the other
tests pass.  Added that check to the _stop() function.  Also renamed it to
the more explanatory _device_not_valid().


geoffl:
looks good

PR #384, opened 2012-07-08
bz19345: Completely remove delayed sharing start feature.

geoffl:
Don't delay the sharing start in any way until we think about this further.
I think what's happened is we have tried to delete something before the
sharing's started and so it tries to remove an item which the sharing
manager did not know about.

Since sharing wasn't really designed with that in mind let's just disable it
for now since this is the safest.

(note: don't forget to cherry-pick for 5.0.2 if it looks good!)


bendk:
Looks fine to me.

PR #385, opened 2012-07-09
Build with FFmpeg 0.11

ilpianista:
This fixes all build failures with FFmpeg < 0.11.

It breaks the build with previous FFmpeg versions.


bendk:
Hi bashinx, thanks for the pull request.  What happens when this code runs on a system with ffmpeg >= 0.11?  If it won't work, then I think we need to use conditional compilation or something to avoid the conflicts.

ilpianista:
I didn't try (nor I can, as on Arch Linux we already switched to 0.11). You've two choice here:
1) use conditional compilation
2) bump minimum ffmpeg version required

bendk:
I think this code is good, but let's merge into master rather than the 5.0 branch.  I don't think we are going to be releasing any more versions there unless we have to and it seems this isn't a deal-breaker for any distros.

Here's the pull-request for the master branch.
https://github.com/pculture/miro/pull/412

PR #387, opened 2012-07-12
Fix19244

bendk:
This pull contains several changes to prevent us from keeping open file handles on devices when we shouldn't.


geoffl:
other than the popen close_fds thing for windows I think it looks okay.

geoffl:
other than the popen close_fds thing for windows I think it looks okay.

bendk:
Geoff says that I can merge this one myself.  Going for it now.

PR #388, opened 2012-07-12
bz19276 comment #5: fix layout of the 7digital links.

bendk:
Not sure why things worked okay on GTK, but we were packing the HBox wrong.


geoffl:
looks ok

PR #394, opened 2012-07-27
Made ubuntu helper scripts executable

bendk:


bendk:
This pull request is now more than just a chmod, but I'm too lazy to open up a new one.  It's pretty trivial in any case.

PR #395, opened 2012-07-28
Added an extra commit/begin pair in upgrade189.

bendk:
Otherwise this upgrade crashes on ubuntu oneric (and presumably other
distros with older sqlite versions)


geoffl:
looks good

PR #396, opened 2012-07-28
Fixed SQL triggers that insert values into item_fts

bendk:
If the user upgraded their database, we weren't copying the parent_title text
over.

Added a unittest to check this in the future.


geoffl:
looks good

PR #397, opened 2012-08-02
bz19415: Update binary kit.

geoffl:
Build with binary kit with updated libtorrent and boost to fix
magnet downloads on OS X 10.8.


geoffl:
merging

PR #399, opened 2012-08-03
Fixing the unittests on osx.

bendk:
We don't need to run the info list test anymore since it's being replaced.
Also the itemrenderer test was only working on GTK so let's skip it on OSX.


geoffl:
let's merge

PR #400, opened 2012-08-03
Better way to determine if we create a new database.

bendk:
In the old code, backend would check if the database path didn't exist.  But
if the frontend opened the database first this would fail.


geoffl:
let's merge

bendk:
Merging.

PR #401, opened 2012-08-04
bz19413: Make ItemInfo.thumbnail a Filename when it comes from a screenshot

bendk:
Refactored the various path columns to try to make them more consistent.  Now
data straight from the database is <name>_path_unicode and the data
as a FilenameType is <name>_path

The one exception is filename, which is referenced in a bunch of places, so I
didn't want to change it.  Still, I renamed the corresponding attribute to
filename_unicode.


geoffl:
merging

PR #402, opened 2012-08-08
Osxlayout

bendk:
This is one fix that changes how we lay out things on OSX, then a fix for 19411 (and also 18863 in the process)


geoffl:
merging

PR #405, opened 2012-08-12
bz19441: Relink the py2app application stub with Cocoa and AppKit.

geoffl:
Seems that the sandbox has some trouble finding sandbox services
unless these libraries are dynamically linked in right from the
start in the main executable of the application bundle.


bendk:
I'm not sure what's going on here exactly, but it seems pretty safe to me.

PR #406, opened 2012-08-14
Clean up the scroller layout code a bit.

geoffl:
Clean up the size calculation for the overlay and legacy scroller styles.


bendk:
Looks good.

PR #407, opened 2012-08-15
bz19464: Fix deprecated messages.

geoffl:
Some APIs have been deprecated but they are particularly obnoxious
with spamming the console with deprecation messages.  In the
interest of mitigating this the said APIs have been migrated to
new style APIs.


bendk:
Seems good.

PR #409, opened 2012-08-18
Fix build warnings

marciof:
I've made some changes here and there to fix various build warnings.


bendk:
Looks great to me.  Let's merge.

PR #410, opened 2012-08-29
Devicedata

bendk:


geoffl:
Had a quick look, didn't run it but everything looked more or less ok.  I guess we'll just merge then fix any fallout.

PR #412, opened 2012-08-31
Update segmenter code to not use deprecated functions

bendk:
Took diff from Andrea Scarpino and applied it to master.

This works on ubuntu 12.04, I think it should work on most recent OSes as well.  I'm not really sure though, what do you think Geoff?  If we have to, we can use conditional compilation to make it work for both versions, but I'd rather not unless we had to.

### DIFF
--- a/tv/linux/miro-segmenter.c
+++ b/tv/linux/miro-segmenter.c
@@ -156,7 +156,7 @@ int main(int argc, char **argv)
         exit(1);
     }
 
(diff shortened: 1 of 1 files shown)


geoffl:
I guess this would not work when we recompile on windows/mac since I think we are still on ffmpeg there.

but since it is not used at this moment maybe it's okay.  and when we need to we can deprecate the use of ffmpeg or update to whatever's the latest.

geoffl:
let's just merge

PR #421, opened 2013-01-25
SQL Unicode problems

k-nut:
When connecting my device (with German titles on it) miro would always crash on me with the following exception: https://gist.github.com/4635745
I simply converted the values to unicode and do not get the error anymore. In addition there don't seem to be any unwanted side effects so this seems to be an easy fix.
Your comments on this are very welcome!


bendk:
Thanks for the bug report and also the patch.  I'm not going to merge that code exactly, but I used it + the bug report to make this change:
 https://github.com/pculture/miro/commit/f07e652c0a746e6a1737b00149ed31bdf493c377

I think it should fix the problem you're seeing.  If not, please leave a comment here:
http://bugzilla.pculture.org/show_bug.cgi?id=19815

PR #422, opened 2013-02-05
Add helper script for Fedora 16

ndim:
The difference to fedora_15.sh is to also install sqlite-devel.


ndim:
I cannot see how [the above Travis build failure](https://travis-ci.org/pculture/miro/builds/4597424) (Python 2.6 build works, Python 2.7 failes with 10 minute build job timeout) could be related to this pull request adding a shell script which is not even run during a Travis build.

bendk:
I'm not sure why travis is failing all the time either, but I'm sure it's not related to the helper script.  Let's merge.

PR #428, opened 2014-03-17
Fix YouTube downloads not working

MitchellMcKenna:
- Fix youtube videos not downloading, error in the log said key 'sig'
  was not defined.
- Youtube automatically puts the signature in the 'url' param now, not
  as an extra parameter any longer, so 'sig' param is no longer defined.


warnew:
thanks!

jdragojevic:
This is available on the nightly builds on this page: http://nightlies.pculture.org.  I noticed that it's the .flv file that's getting downloaded by default now.

